PR #1, opened 2022-02-23
Add mtls termination

sonnysideup:


deployments/helm/hephaestus/charts/buildkit/templates/statefulset.yaml
```diff
@@ -16,7 +16,8 @@ spec:
   template:
     metadata:
       annotations:
-        {{- if .Values.rootless }}
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- if and (not (include "hephaestus.supportsSeccompGA" .)) .Values.rootless }}
```
steved, line 20: I'm surprised you can reach back to the parent chart? Feels a bit dirty?

steved (approved):

deployments/helm/hephaestus/charts/buildkit/templates/statefulset.yaml
sonnysideup: I agree, this is filthy schmutz, but according to the Helm [subchart docs](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#sharing-templates-with-subcharts):

> Parent charts and subcharts can share templates. Any defined block in any chart is available to other charts.

🤷🏼‍♂️🤷🏼‍♂️🤷🏼‍♂️

sonnysideup: I may ultimately just subsume the buildkit subchart into the parent chart under a `buildkit` directory. Dunno yet, open to suggestions.

PR #2, opened 2022-03-29
Implements post-processing of controller logs

sonnysideup:
- adds logging multiplexer
- refactors helm chart values
- modifies controller config to allow for logging and buildkit adv cfg
- removes api fields no longer used

### DIFF
--- a/deployments/helm/hephaestus/values.yaml
+++ b/deployments/helm/hephaestus/values.yaml
@@ -1,106 +1,172 @@
-# Buildkit deployment configuration.
-# See charts/buildkit/values.yaml for a list of available values.
-buildkit:
-  replicaCount: 1
-  mtls:
-    enabled: true
+# Global parameters applied to all images in this chart
+global:
+  imageRegistry: ""
+  imagePullSecrets: []
 
-# Number of controller instances to run. Leader election will be enabled
-# whenever this value is greater than 1; this ensures that only one instance
-# is active at a time.
-replicaCount: 1
+# Replace the name of the chart defined in Chart.yaml
+nameOverride: ""
 
-config:
-  # Webhook server port
-  webhookPort: 9443
-  # Prometheus metric ports
-  metricsPort: 8080
-  # Health probe port
-  healthProbePort: 8081
-  # Limit watch to a specific set of namespaces, default is all namespaces
-  watchNamespaces: []
+# Replace the generated name (release+chart name)
+fullnameOverride: ""
 
-  # Logger encoding can be either 'json' or 'console'
-  logEncoder: ""
-  # Logger level can be either 'debug', 'info', 'error', or any integer value >= 0 which corresponds to custom debug levels of increasing verbosity
-  logLevel: ""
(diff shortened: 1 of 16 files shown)


deployments/helm/hephaestus/values.yaml
```diff
@@ -1,106 +1,172 @@
...
 
-nameOverride: ""
-fullnameOverride: ""
+  # Resource requests and limits
+  resources:
+    requests: {}
+    limits: {}
+
+  # Configure the ingestion, transformation, and delivery of your logs
+  config:
```
steved, line 143: It does feel a tiny bit brittle if we're asking e.g. our deployments to configure this with the helm templating included, but I might need a more concrete example of how we want to configure sinks before balking at it. For example, maybe we'll never need to template `sinks` and we're better off with it being actual structured input?

pkg/logger/logger.go
```diff
@@ -2,48 +2,94 @@ package logger
...
+		if err != nil {
+			return logr.Logger{}, fmt.Errorf("cannot create logfile logger: %w", err)
 		}
-		opts = append(opts, ctrlzap.StacktraceLevel(lvl))
+
+		level, err := parseLevel(cfg.Logfile.LogLevel)
+		if err != nil {
+			return logr.Logger{}, fmt.Errorf("invalid logfile log level: %w", err)
+		}
+		fileCore := zapcore.NewCore(&ctrlzap.KubeAwareEncoder{Encoder: jsonEncoder}, zapcore.Lock(file), level)
```
steved, line 55: Do we need to consider [file rotation](https://github.com/uber-go/zap/blob/master/FAQ.md#does-zap-support-log-rotation)? Feel like we chatted about it, but I don't recall what the outcome was

steved (approved):
Couple comments, but LGTM so far.

deployments/helm/hephaestus/values.yaml
sonnysideup: My thinking is that both the S3 and Redis sinks have a number of advanced configuration options available that I don't want to constrain nor map out at a higher level. Additionally, other sinks seems interesting (Kafka, NewRelic, etc...) and it might be worth leaving the `sink` config free-form so that it can be extended w/o modifications to the helm chart. That was my reasoning, thoughts?

https://vector.dev/docs/reference/configuration/sinks/vector/

pkg/logger/logger.go
sonnysideup: This totally slipped my mind. I think we _should_ control the logfile size but I'll need to test what, if any, extra configuration is required by Vector to make sure it's not stuck on an old file inode.

sonnysideup: It looks like the default `file` source from Vector automatically picks up a file rotation event and starts tailing the new logfile.

```
hephaestus-manager-58dfdff677-pkmtq log-processor 2022-03-30T16:59:18.907340Z  INFO source{component_kind="source" component_id=hephaestus_logs component_type=f
ile component_name=hephaestus_logs}:file_server: vector::internal_events::file::source: Found new file to watch. file=/var/log/hephaestus/output.json
```

PR #3, opened 2022-04-01
Convert ContainerImageBuild to ImageBuild resources

sonnysideup:
Adds a basic controller to convert `forgev1alpha1.ContainerImageBuild` into `hephv1.ImageBuild`resources. This is intended to aid our efforts to quickly migrate from Forge to Heph for image building.


pkg/api/forge/v1alpha1/containerimagebuild_types.go
```diff
@@ -0,0 +1,226 @@
...
+	// +kubebuilder:validation:Optional
+	PluginData map[string]string `json:"pluginData"`
+
+	// Disable the use of layer caches during build.
+	// +kubebuilder:validation:Optional
+	DisableBuildCache bool `json:"disableBuildCache"`
+
+	// Disable export of layer cache when it is enabled.
+	// +kubebuilder:validation:Optional
+	DisableLayerCacheExport bool `json:"disableLayerCacheExport"`
```
steved, line 159: Do we need to track parity for a few of these? `DisableLayerCacheExport`, `DisableBuildCache`, `ImageSizeLimit`?

steved (approved):

pkg/api/forge/v1alpha1/containerimagebuild_types.go
sonnysideup: How do `DisableLayerCacheExport` and `DisableBuildCache` work in V1? This model somewhat mimics that design with its localized disk cache so I'm not sure how to remove that but I can look.

I figured out how to implement `ImageSizeLimit` with https://github.com/dominodatalab/hephaestus/blob/67c91d24ea7e7f6ce64a4eb7cb3bf3e513157a12/pkg/buildkit/buildkit.go#L194-L195. It's just a question if we really want it. What do you think?

sonnysideup: I think `DisableLayerCacheExport` ( push cache to registry ) made sense in Forge but doesn't really have meaning in either V1 or V3. I think we do not need to worry about this, do you disagree?

`DisableBuildCache` can be implemented with the use of the `no-cache` attribute when performing a build. This will ensure a single build does not use the stored (filesystem) cache for any of its steps.

steved: > I think DisableLayerCacheExport ( push cache to registry ) made sense in Forge but doesn't really have meaning in either V1 or V3. I think we do not need to worry about this, do you disagree?

Is this maybe confusion over terminology? Are we not going to need [instruction caching at all](https://github.com/moby/buildkit/blob/master/README.md#inline-push-image-and-cache-together)? 

> DisableBuildCache can be implemented with the use of the no-cache attribute when performing a build. This will ensure a single build does not use the stored (filesystem) cache for any of its steps.

+1, but does the image build resource need a field for it?

sonnysideup: I've added these fields and I'm going to implement this support in another PR.

PR #4, opened 2022-04-05
CRD Apply/Delete

sonnysideup:


pkg/crd/crd.go
```diff
@@ -58,17 +58,26 @@ var (
...
 // operate will read all available CRDS and apply state changes to the cluster using the processor func.
-func operate(ctx context.Context, processor crdProcessor) error {
+func operate(ctx context.Context, processor crdProcessor, istio bool) error {
+	if istio {
+		quit, err := waitForIstioSidecar()
+		if err != nil {
+			return err
+		}
+
+		defer quit()
```
steved, line 78: You need `quit` to run regardless of `err`, no? Otherwise it'll just hang?

steved (approved):

pkg/crd/crd.go
sonnysideup: Hmm, is that true? The only reason I can see an `err` occurring here is if we fail to establish the Istio sidecar's health in the first place. If we can't assert Istio's health, can we then tell Istio to quit?

steved: Fair point

PR #5, opened 2022-04-08
Adds anti-affinity term to buildkit pods

sonnysideup:


deployments/helm/hephaestus/charts/buildkit/values.yaml
```diff
@@ -65,7 +65,22 @@ nodeSelector: {}
...
+            matchExpressions:
+              - key: app.kubernetes.io/instance
+                operator: In
+                values:
+                  - {{ .Release.Name }}
+              - key: app.kubernetes.io/name
+                operator: In
+                values:
+                  - {{ .Chart.Name }}
+          topologyKey: "kubernetes.io/hostname"
```
steved, line 83: Can we add another on AZ with a lower weight?

steved (approved):

deployments/helm/hephaestus/charts/buildkit/values.yaml
steved: Yep, thanks!

PR #6, opened 2022-04-08
Preparing artifacts for distribution

sonnysideup:
i don't know if a "composite action", "reusable workflow", or duplicated
steps is the best approach here. this is the first attempt.


sonnysideup:
@steved Pushing images to GHCR/Quay and Helm chart to GHCR/GCR. Using the composite action seemed like the best option considering all others.

steved (approved):

PR #8, opened 2022-04-13
Polish Helm chart

sonnysideup:
- Condenses Helm chart (i.e. removes buildkit subchart)
- Adds network policies
- Adds pod security policies
- Tests with Istio (w/ and w/o CNI plugin) v1.13.2
- Executes vector container as non-root user
- Converts controller container back to non-root user


deployments/helm/hephaestus/templates/_helpers.tpl
```diff
@@ -36,35 +45,60 @@ Return the service account name.
...
 {{- end }}
 
 {{/*
-Returns the logfile directory.
+Return whether or not securityContext.seccompProfile field is supported.
 */}}
-{{- define "hephaestus.logfileDir" -}}
-/var/log/hephaestus
+{{- define "hephaestus.supportsSeccompGA" -}}
+{{- semverCompare ">1.19-0" .Capabilities.KubeVersion.Version }}
```
fraenkel, line 55: with 5.2, we only support v1.19 and higher so its your call

deployments/helm/hephaestus/templates/_helpers.tpl
```diff
@@ -8,21 +8,30 @@ Return the proper controller image name.
...
 
 {{/*
-Return the proper log processor image name.
+Return the controller standard labels.
 */}}
-{{- define "hephaestus.logprocessor.image" -}}
-{{- include "common.images.image" (dict "imageRoot" .Values.logProcessor.image "global" .Values.global) }}
+{{- define "hephaestus.controller.labels.standard" -}}
+{{- include "common.labels.standard" . }}
+app.kubernetes.io/component: controller
```
fraenkel, line 15: we use `    {{- include "common.labels.component" "cainjector" | nindent 4 }}`

deployments/helm/hephaestus/templates/controller/clusterrole.yaml
```diff
@@ -74,3 +74,11 @@ rules:
       - get
       - watch
       - patch
+  {{- if and (include "hephaestus.pspRequired" .) (include "hephaestus.istioWithoutCNI" .) }}
```
fraenkel, line 77: why are both required?

deployments/helm/hephaestus/values.yaml
```diff
@@ -13,14 +13,20 @@ fullnameOverride: ""
...
+
 # Istio configuration
 istio:
   # Enable support for environments with Istio installed
   enabled: false
-  # NOTE: check if this setting is necessary (i.e. run with Istio with CNI and PSPs)
   # Elevate pod execution permissions so that Istio's init container can modify
   # network settings when CNI plugin is NOT installed.
-  cniPluginInstalled: true
+  cniPluginInstalled: false
```
fraenkel, line 29: all our other charts just use `cni: true/false`

fraenkel (changes requested):

deployments/helm/hephaestus/templates/_helpers.tpl
sonnysideup: that's good to know. I feel fine removing this then.

deployments/helm/hephaestus/templates/_helpers.tpl
sonnysideup: I've been using the [bitnami common](https://github.com/bitnami/charts/tree/master/bitnami/common#labels) library chart for a while now and, unfortunately, they don't have that function available.

I _believe_ I wrote this function way back when and it does essentially the same thing.

deployments/helm/hephaestus/values.yaml
sonnysideup: I'm fine changing the field name here. 👍🏼

deployments/helm/hephaestus/templates/controller/clusterrole.yaml
sonnysideup: Good question, here's the answer:

You already know Istio w/o CNI means every Pod that is launched requires extra CAPs. and a root user. Hence, we need to do a little extra work whenever PSPs are enabled and configured to be _strict_.

The PSP API will be removed in k8s 1.25 so `include "hephaestus.pspRequired" .` macro will ensure that we don't render them whenever we start deploying to that version or greater. Without it, we'll get errors when trying to create resources for undefined object APIs and, in the case of RBAC objects, create superfluous rules that might confuse people

fraenkel: I know why its needed with CNI, but given we have enablePSPs as a separate thing it should be an OR not an AND. PSPs are there for other things as well.
I agree the additional CAPs should be guarded by CNI but not the rest.

fraenkel (approved):

deployments/helm/hephaestus/templates/controller/clusterrole.yaml
sonnysideup: changed the condition.

PR #10, opened 2022-05-06
add an example skaffold setup, cache golang builds

steved:


steved:
https://github.com/GoogleContainerTools/skaffold/issues/7284 is annoying

PR #11, opened 2022-05-07
allowing providing secrets to hephaestus to pass to builds

steved:
e.g.

```yaml
controller:
  extraVolumes:
    - name: cert-vol
      configMap:
        name: domino-generated-certificates
  extraVolumeMounts:
    - name: cert-vol
      mountPath: /certs
  secrets:
    cacerts.p12: "/certs/cacerts.p12"
    ca-certificates.crt: "/certs/ca-certificates.crt"
```

and referencing them:
```dockerfile
FROM ...
RUN --mount=type=secret,required=true,target=/etc/ssl/certs/ca-certificates.crt,mode=444 \
  --mount=type=secret,required=true,target=/etc/ssl/certs/cacerts.p12,mode=444 \
  ls -la /etc/ssl/certs/
```

### DIFF
--- a/deployments/helm/hephaestus/values.yaml
+++ b/deployments/helm/hephaestus/values.yaml
@@ -88,6 +88,9 @@ controller:
   # Defaults to "10m"
   poolMaxIdleTime: null
 
+  # Secrets (name: path) to expose into builds that request it
+  secrets: {}
+
   # Resource requests and limits
   resources:
     requests: {}
@@ -126,6 +129,11 @@ controller:
 
   # Manager pods priorityClassName
   priorityClassName: ""
+
+  # Additional volume mounts
+  extraVolumeMounts: []
+  # Additional volumes
+  extraVolumes: []
 
   # Number of manager instances to run. Leader election will be enabled
   # whenever this value is greater than 1; this ensures that only one instance
--- a/pkg/buildkit/buildkit.go
+++ b/pkg/buildkit/buildkit.go
@@ -14,6 +14,7 @@ import (
 	bkclient "github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/cmd/buildctl/build"
 	"github.com/moby/buildkit/session"
+	"github.com/moby/buildkit/session/secrets/secretsprovider"
 	"github.com/moby/buildkit/util/progress/progressui"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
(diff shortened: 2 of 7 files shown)


pkg/buildkit/buildkit.go
```diff
@@ -38,6 +40,11 @@ func (b *clientBuilder) WithDockerAuthConfig(configDir string) *clientBuilder {
 	return b
 }
 
+func (b *clientBuilder) WithSecrets(secrets map[string][]byte) *clientBuilder {
+	b.secrets = secrets
+	return b
+}
+
```
sonnysideup, line 47: I know we only use a newly-constructed client "once" but do you think this should this be a builder option or a part of the `BuildOptions` passed to `Build` since mounting is build-specific. I only ask because I was thinking about moving the `WithDockerAuthConfig` out of the builder and into the build options since that is less client-specific.

sonnysideup (commented):
Looks good so far. I was thinking vols & secrets would have to be mounted directly into the buildkit containers but using a `session.Attachable` with the secrets totally makes sense.

pkg/buildkit/buildkit.go
steved: I forgot we should be loading the files per-build (since configmaps can change), so I think it's totally reasonable for this to be on `Build`. Is 31dfb37 what you had in mind?

deployments/helm/hephaestus/values.yaml
```diff
@@ -88,6 +88,9 @@ controller:
   # Defaults to "10m"
   poolMaxIdleTime: null
 
+  # Secrets (name: path) to expose into builds that request it
+  secrets: {}
```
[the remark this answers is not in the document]
steved: Yeah, I struggle with this since there's going to be some coupling either way:

* The thing invoking the build has to somehow know what's attached to the hephaestus pod, but that invoker can determine the ID of the secret and pass that into the Dockerfile in a single place.
* Here, we at least keep the mounting of the configmap and reference to it in the same place, but the Dockerfile just uses a "magic" ID that comes from seemingly nowhere.

pkg/buildkit/buildkit.go
sonnysideup: > Is https://github.com/dominodatalab/hephaestus/commit/31dfb371e0233606f6a3dc59dcb1c15f21fb18a3 what you had in mind?

Yes sir.

deployments/helm/hephaestus/values.yaml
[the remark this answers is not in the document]
sonnysideup: That is a bit of a pickle, isn't it? I think this will suffice for the first cut and we can figure the final solution in the near future like we discussed offline today.

sonnysideup (approved):

PR #15, opened 2022-05-10
Cancel buildkitd Solve on ImageBuild delete

sonnysideup:
- separate context handling when build a new buildkit client and calling `Solve` on that client
- adds `DeleteBroadcaster` to watch for and emit messages when image build resources are deleted
- `BuildDispatcher` listens for and acts upon delete broadcasts to cancel buildkit `Solve` contexts.

### DIFF
--- a/pkg/buildkit/buildkit.go
+++ b/pkg/buildkit/buildkit.go
@@ -78,7 +78,7 @@ type BuildOptions struct {
 
 type Buildkit interface {
 	Build(ctx context.Context, opts BuildOptions) error
-	Cache(image string) error
+	Cache(ctx context.Context, image string) error
 }
 
 type Client struct {
@@ -199,8 +199,8 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 	return c.runSolve(ctx, solveOpt)
 }
 
-func (c *Client) Cache(image string) error {
-	return c.solveWith(context.TODO(), func(buildDir string, solveOpt *bkclient.SolveOpt) error {
+func (c *Client) Cache(ctx context.Context, image string) error {
+	return c.solveWith(ctx, func(buildDir string, solveOpt *bkclient.SolveOpt) error {
 		dockerfile := filepath.Join(buildDir, "Dockerfile")
 		contents := []byte(fmt.Sprintf("FROM %s\nRUN echo extract\n", image))
 		if err := os.WriteFile(dockerfile, contents, 0644); err != nil {
@@ -267,7 +267,10 @@ func (c *Client) runSolve(ctx context.Context, so bkclient.SolveOpt) error {
 			c = cn
 		}
 
-		_, err := progressui.DisplaySolveStatus(context.TODO(), "", c, lw, ch)
+		// this operation should return cleanly when solve returns (either by itself or when cancelled) so there's no
+		// need to cancel it explicitly. see https://github.com/moby/buildkit/pull/1721 for details.
+		_, err := progressui.DisplaySolveStatus(context.Background(), "", c, lw, ch)
+
 		return err
 	})
 
--- a/pkg/buildkit/buildkit.go
+++ b/pkg/buildkit/buildkit.go
@@ -22,15 +22,14 @@ import (
 )
 
 type clientBuilder struct {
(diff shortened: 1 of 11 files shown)


pkg/buildkit/buildkit.go
```diff
@@ -199,11 +196,11 @@ func (c *Client) Build(opts BuildOptions) error {
 	}
 
 	// build/push images
-	return c.runSolve(solveOpt)
+	return c.runSolve(ctx, solveOpt)
 }
 
 func (c *Client) Cache(image string) error {
-	return c.solveWith(func(buildDir string, solveOpt *bkclient.SolveOpt) error {
+	return c.solveWith(context.TODO(), func(buildDir string, solveOpt *bkclient.SolveOpt) error {
```
steved, line 203: why not pass this context through the caller even though it's unused?

steved, line 203: why not pass this through the caller even though it's unused?

pkg/controller/imagebuild/component/deletebroadcaster.go
```diff
@@ -0,0 +1,46 @@
...
+		return ctrl.Result{}, err
+	}
+
+	log.Info("Broadcasting delete message")
+	c.delete <- obj.ObjectKey()
+
+	return ctrl.Result{}, nil
+}
+
+// ctx, cancel := context.WithCancel(coreCtx)
```
steved, line 41: ✂️ ?

steved, line 41: ✂️ ?

pkg/controller/imagebuild/component/builddispatcher.go
```diff
@@ -44,13 +53,22 @@ func (c *BuildDispatcherComponent) Initialize(ctx *core.Context, _ *ctrl.Builder
...
 }
 
 func (c *BuildDispatcherComponent) Reconcile(ctx *core.Context) (ctrl.Result, error) {
 	log := ctx.Log
 	obj := ctx.Object.(*hephv1.ImageBuild)
 
+	buildCtx, cancel := context.WithCancel(ctx)
+	c.cancels.Store(obj.ObjectKey(), cancel)
+	defer func() {
+		cancel()
```
steved, line 68: do you need to explicitly call `cancel`?

steved, line 68: do you need to explicitly call `cancel`?

pkg/buildkit/buildkit.go
```diff
@@ -256,31 +253,31 @@ func (c *Client) solveWith(modify func(buildDir string, solveOpt *bkclient.Solve
...
-		if _, err := c.bk.Solve(ctx, nil, so, ch); err != nil {
-			return err
+		var c console.Console
+		if cn, err := console.ConsoleFromFile(os.Stderr); err != nil {
+			c = cn
 		}
 
-		c.log.Info("Solve complete")
-		return nil
+		_, err := progressui.DisplaySolveStatus(context.TODO(), "", c, lw, ch)
```
steved, line 270: Why not `ctx` here like before?

sonnysideup: Great question. That was my original design but I noticed that sometimes the build would hang indefinitely after cancellation. A little digging revealed that, if the progress printer is cancelled before solve returns, then a routine inside the latter will block when writing to the status channel since there is nothing left to read from it and the `Wait()` func would hang indefinitely.

Details here: https://github.com/moby/buildkit/pull/1721

pkg/buildkit/buildkit.go
sonnysideup: makes sense, i was just being lazy.

pkg/buildkit/buildkit.go
sonnysideup: i added a quick change and note to indicate the reason why this was done. what do you think?

pkg/controller/imagebuild/component/builddispatcher.go
steved: TIL we should always explicitly cancellable contexts, makes sense though!

steved (approved):

PR #19, opened 2022-05-13
Adds separate build job to create custom buildkit image

sonnysideup:


build/buildkit/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG BUILDKIT_TAG=master-rootless
+
+FROM moby/buildkit:${BUILDKIT_TAG}
+ARG ISTIO_GID=1337
+USER root
+RUN addgroup -S -g $ISTIO_GID istio && \
```
steved, line 6: Can you run this as well to ensure we get the updated alpine packages?

```


RUN apk update && \
  apk upgrade && \
  rm -rf /var/cache/apk
```

steved (approved):

build/buildkit/Dockerfile
sonnysideup: Tested a couple of builds after the `apk upgrade` and the image building is unaffected.

PR #20, opened 2022-05-16
Adds logProcessor.extraVolumeMounts value to chart

sonnysideup:
So that you could hypothetically add volumes and vector configurations to write files into them.

### DIFF
--- a/deployments/helm/hephaestus/values.yaml
+++ b/deployments/helm/hephaestus/values.yaml
@@ -132,6 +132,7 @@ controller:
 
   # Additional volume mounts
   extraVolumeMounts: []
+
   # Additional volumes
   extraVolumes: []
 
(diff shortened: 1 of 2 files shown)


deployments/helm/hephaestus/values.yaml
```diff
@@ -181,6 +182,10 @@ logProcessor:
         encoding:
           codec: json
 
+  # Additional volume mounts into log processor container. Volumes are listed
+  # in controller.extraVolumes since the two containers live in the same pod.
```
steved, line 186: Feel free to move extraVolumes to the top-level, I forgot it's shared.

steved (approved):

deployments/helm/hephaestus/values.yaml
sonnysideup: I saw that and it somewhat begs the question, "Is this value layout correct?" Top-level might-could be mistaken to refer to buildkit, right?

Probs should be:

```yaml
# truly global values like enableNetworkPolicies and global.imageRegistry

controller:
  manager:
    ...
  logProcessor:
    ...
buildkit:
  ...
```

sonnysideup: Added https://github.com/dominodatalab/hephaestus/issues/21 for a 2nd-sweep. I'm fine with this for now.

PR #22, opened 2022-05-17
Enables multi-namespace helm chart deploy

sonnysideup:
Allows users to deploy controller and buildkit workloads into separate namespaces.

### DIFF
--- a/deployments/helm/hephaestus/templates/buildkit/networkpolicy.yaml
+++ b/deployments/helm/hephaestus/templates/buildkit/networkpolicy.yaml
@@ -23,6 +23,6 @@ spec:
           {{- if ne (include "hephaestus.buildkit.namespace" .) .Release.Namespace }}
           namespaceSelector:
             matchLabels:
-              {{- toYaml (lookup "v1" "Namespace" "" .Release.Namespace).metadata.labels | nindent 14 }}
+              {{- .Values.buildkit.releaseNamespaceLabels | default (lookup "v1" "Namespace" "" .Release.Namespace).metadata.labels | toYaml | nindent 14 }}
           {{- end }}
 {{- end }}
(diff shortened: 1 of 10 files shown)


deployments/helm/hephaestus/templates/buildkit/networkpolicy.yaml
```diff
@@ -19,4 +20,9 @@ spec:
         - podSelector:
             matchLabels:
               {{- include "hephaestus.controller.labels.matchLabels" . | nindent 14 }}
+          {{- if ne (include "hephaestus.buildkit.namespace" .) .Release.Namespace }}
+          namespaceSelector:
+            matchLabels:
+              {{- toYaml (lookup "v1" "Namespace" "" .Release.Namespace).metadata.labels | nindent 14 }}
```
steved, line 26: 🤯 

This concerns me a bit if labels change, though I don't know if I have any evidence of that.

sonnysideup: I could see a situation where ` (lookup "v1" "Namespace" "" .Release.Namespace).metadata.labels` returns something like:

```yaml
# these don't change
kubernetes.io/metadata.name: <ns-name>
dominodatalab.com/node-pool: <bleh>

# however these could
custom-customer-label: foo
silly-openshift-metadata: v1.3.2.5.23423423
...
```

Adding labels is a non-issue but if users intentionally or inadvertently change the values of existing labels, then this shit breaks. We may want to offer the ability to override the selector to focus on a single k/v pair, right?

steved: Right, we have static labels like `domino-platform`, but those custom labels were my concern. An override couldn't hurt.

PR #30, opened 2022-05-24
Builds out new buildkit workerpool implementation

sonnysideup:
a number of separate concerns have been broken out into separate types that focus on a single aspect of "scaling workers".


pkg/buildkit/worker/leasemanager.go
```diff
@@ -0,0 +1,345 @@
...
+	if !errors.Is(err, errNoUnleasedPods) {
+		return nil, fmt.Errorf("unexpected leasing error: %w", err)
+	}
+
+	m.log.Info("Scaling cluster, no unleased pods found")
+	if err = m.scaler.Scale(ctx, 1); err != nil {
+		return nil, err
+	}
+
+	m.log.Info("No unleased pods found, scaling workload")
```
steved, line 209: I'm not sure I understand this log message in this context

pkg/buildkit/worker/leasemanager.go
```diff
@@ -0,0 +1,345 @@
...
+	if pod, err = m.getUnleasedPod(ctx); err != nil {
+		return nil, fmt.Errorf("unable to acquire lease after scaling: %w", err)
+	}
+
+	return pod, nil
+}
+
+func (m *podLeaseManager) getUnleasedPod(ctx context.Context) (*corev1.Pod, error) {
```
steved, line 217: Is there a race condition here if multiple callers come through `nextAvailablePod`? E.g.

* two pods trigger a scale up at the same time
* the list returns two unleased pods with each goroutine capturing the first unleased pod
* that pod is then passed twice into `ptch`?

Maybe this just needs an `mu` lock?

pkg/buildkit/worker/leasemanager.go
```diff
@@ -0,0 +1,345 @@
...
+addressScan:
+	for _, subset := range endpoints.Subsets {
+		for _, address := range subset.Addresses {
+			if address.TargetRef.Name == podName {
+				for _, port := range subset.Ports {
+					if port.Port == m.servicePort {
+						host = strings.Join([]string{address.Hostname, endpoints.Name, endpoints.Namespace}, ".")
+						m.log.Info("Found eligible endpoint address", "host", host)
+
+						break addressScan
```
steved, line 253: This just screams separate method that takes endpoints and and has a `return host` to me. I had to lookup label breaks again.

pkg/buildkit/worker/leasemanager.go
```diff
@@ -0,0 +1,345 @@
...
+	defer m.mu.Unlock()
+
+	m.log.Info("Querying for pods", "namespace", m.namespace, "opts", m.podListOptions)
+	list, err := m.podClient.List(ctx, m.podListOptions)
+	if err != nil {
+		return err
+	}
+
+	var removals []string
+	for idx := len(list.Items) - 1; idx >= 0; idx-- {
```
steved, line 303: Are the pods sorted ordinally?

pkg/buildkit/worker/leasemanager.go
```diff
@@ -0,0 +1,345 @@
...
+		pod := list.Items[idx]
+		log := m.log.WithValues("podName", pod.Name)
+
+		if id, ok := pod.Annotations[managerIDAnnotation]; ok && id != m.uuid {
```
steved, line 307: When would the manager ID annotation exist and be different than expected? Or just defensive coding?

pkg/buildkit/worker/leasemanager.go
```diff
@@ -0,0 +1,345 @@
...
+			continue
+		}
+
+		if _, leased := pod.Annotations[leasedAnnotation]; leased {
+			break
+		}
+
+		ts, ok := pod.Annotations[expiryTimeAnnotation]
+		if !ok {
+			m.log.Info("Eligible for termination, missing expiry time")
```
steved, line 320: May want to log the pod name (and probably in each of the log lines in the loop)?

steved (approved):

pkg/buildkit/worker/leasemanager.go
sonnysideup: fixed in https://github.com/dominodatalab/hephaestus/pull/30/commits/7f6c17bdcd6db509f8ded5e5550c88729de22d36

pkg/buildkit/worker/leasemanager.go
sonnysideup: whoops, good catch. i meant to remove this.

pkg/buildkit/worker/leasemanager.go
sonnysideup: It was a little difficult to think about leasing/unleasing in the scope of controller restarts so this is how I accomplished it.

### Metadata

Manager ID is added in `Lease` and removed in `Release.

https://github.com/dominodatalab/hephaestus/blob/7f6c17bdcd6db509f8ded5e5550c88729de22d36/pkg/buildkit/worker/leasemanager.go#L120-L123

https://github.com/dominodatalab/hephaestus/blob/7f6c17bdcd6db509f8ded5e5550c88729de22d36/pkg/buildkit/worker/leasemanager.go#L160-L164

### Predicates

If the controller restarts for some reason (terminated manually, by k8s, or upgrade), then we need to ensure the pods are cleaned up:

1. Pods that were released will NOT have a manager ID and the worker pool GC routine will honor their expiry time.
2. Pods that were leased will have both a lease and manager ID annotation. Since the new lease manager instance will have a new UUID, it will mark these for termination.
3. Pods with invalid metadata (no manager ID, no expiry time) will be marked for termination.

sonnysideup: Does this make sense now? Would you change anything?

steved: I didn't think of this case:
> Pods that were leased will have both a lease and manager ID annotation. Since the new lease manager instance will have a new UUID, it will mark these for termination.

That makes sense, we can't track builds after a restart.

pkg/buildkit/worker/leasemanager.go
sonnysideup: I did some testing with non-buildkit pods (different pod ages, names, workload types) and every invocation of the `List` func produced a slice of items sorted by name in ascending order. Yet, neither the [API docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#list-pod-v1-core) nor the client-go source code indicate a "clear and intentional" sorting order. I did find a reddit post that says

>Someone at kubernetes slack channel answered my question:
>
>api server retrieves resources data from etcd and uses no sorting by default, which defaults to etcd default sorting, which is alphabetical

but that's hardly a credible source.

Despite the fact that the statefulset creates the pods in ascending order, I don't think there's any informantion to rely on this sorting. Furthermore, I can reverse sort this list and then range over it w/o having to use countdown `for` loop.

sonnysideup:
@steved https://github.com/dominodatalab/hephaestus/pull/30#commitcomment-74427291 fixed with https://github.com/dominodatalab/hephaestus/pull/30/commits/3f1f6b027d37428fc4672d30cf0202cec134eff3

pkg/buildkit/worker/leasemanager.go
sonnysideup: @steved check out the updated [worker/pool.go](https://github.com/dominodatalab/hephaestus/blob/b89ea70f7979ac9d0621af9bc41340734f95808c/pkg/buildkit/worker/pool.go). i'll be very surprised if you find a race condition in it.

pkg/buildkit/worker/pool.go
```diff
@@ -0,0 +1,446 @@
...
+	return wp
+}
+
+func (p *workerPool) Get(ctx context.Context) (string, error) {
+	ch := make(chan PodRequestResult, 1)
+	defer close(ch)
+
+	p.requests.Enqueue(ch)
+
+	oob, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
sonnysideup: _out of band_ – it's dumb. do you have a better suggestion?

pkg/buildkit/worker/pool.go
```diff
@@ -0,0 +1,446 @@
...
+
+// async unleased pod request routing
+func (p *workerPool) processRequests() {
+	for {
+		select {
+		case <-p.ctx.Done():
+			close(p.results)
+			for p.requests.Len() > 1 {
+				close(p.requests.Dequeue())
+			}
```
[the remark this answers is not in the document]
sonnysideup: yeah, that works as well.

pkg/buildkit/worker/pool.go
```diff
@@ -0,0 +1,446 @@
...
+	delete(pac.Annotations, managerIDAnnotation)
+
+	p.log.Info("Applying pod metadata changes", "annotations", pac.Annotations)
+	if pod, err = p.podClient.Apply(ctx, pac, metav1.ApplyOptions{FieldManager: fieldManagerName}); err != nil {
+		return fmt.Errorf("cannot update pod metadata: %w", err)
+	}
+
+	if request := p.requests.Dequeue(); request != nil {
+		request <- PodRequestResult{pod, nil}
+	}
```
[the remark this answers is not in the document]
sonnysideup: That's a good question, plus I thought a lot about this. It would be nice to keep dequeuing localized but here's why I do this:

1. You release `pod-N` and send it to `results`.
2. X minutes go pass, the pool is GC'd removing that _unleased_ pod, and that result stays on the channel.
3. Next build receives the stale `pod-N` message from the channel and tries to build.

Is the pod gone? Maybe. It's a race.

pkg/buildkit/worker/pool.go
```diff
@@ -0,0 +1,446 @@
...
+
+	p.log.Info("Scaling cluster, no unleased pods found")
+	if err = p.scaler.Scale(ctx, 1); err != nil {
+		return nil, err
+	}
+	p.log.Info("Scale cluster complete")
+
+	pod, err = p.leasePod(ctx)
+	if err != nil {
```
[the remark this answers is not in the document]
sonnysideup: The current issue I'm facing right now is  dealing with scale up failures on clusters that have a _fixed size_. If we scale up and it fails, then you create a pod that indefinitely stays in a `Pending` state. This matter is compounded with subsequent scale requests.

### Example

1. Your cluster can only run a single buildkit pod allocated with CPU 4 / Memory 16Gi
2. You request 2 builds simulatenously create a `Pending` pod
3. Both requests are eventually serviced by the first buildkit pod
4. You request 2 more builds simultaneously and create another `Pending` pod
5. So forth and so on

I think a loop here will only exacerbate that condition. I think we only try to scale up once, give up if that fails and wait for an unleased pod to make itself available.

PR #33, opened 2022-05-27
update worker pool implementation to use a single update loop for distributing requests

steved:
To simplify the distribution of requests & scaling the cluster, this consolidates the decision making to a single loop. The only downside is that a build request may require multiple `poolSyncTime` passes before it's assigned to an actual pod.

This PR also decreases the default sync time to 30s from 1 minute.


pkg/buildkit/worker/pool.go
```diff
@@ -79,72 +84,85 @@ func NewPool(ctx context.Context, clientset kubernetes.Interface, conf config.Bu
...
 
 	return wp
 }
 
 func (p *workerPool) Get(ctx context.Context) (string, error) {
 	ch := make(chan PodRequestResult, 1)
-	defer close(ch)
 
 	p.requests.Enqueue(ch)
+	defer p.requests.Remove(ch)
```
sonnysideup, line 143: when is this channel closed? the line below

```go
		// when the channel is closed, we receive nil
		if result.pod != nil {
			return p.buildEndpointURL(ctx, result.pod.Name)
		}
```

indicates we need to guard against a close channel, but i don't see anywhere where we close these channels.

pkg/buildkit/worker/pool.go
```diff
@@ -400,37 +321,106 @@ func (p *workerPool) terminateUnleasedPods(ctx context.Context) error {
...
+		},
+		metav1.UpdateOptions{FieldManager: fieldManagerName},
+	)
+	if err != nil {
+		return err
 	}
 
-	p.log.Info("Attempting to terminate pods via scale", "names", removals)
-	return p.scaler.Scale(ctx, count)
+	return nil
```
sonnysideup, line 407: `return err` ?

sonnysideup (approved):
LGTM. 

I guess we're just going to leave follow-up reconciliations that lease pods to the "tick" for now? It seems fine btw.

steved:
> I guess we're just going to leave follow-up reconciliations that lease pods to the "tick" for now? It seems fine btw.

Added in 978448d4d8d07414bee231e7d3af304ed2692874, with the tracked, leased pods it's pretty straightforward.

PR #34, opened 2022-05-28
buildkit: PVC labels are immutable, remove the chart version

steved:
`labels.standard` include the chart version, which means every upgrade of hephaestus will fail because the PVC template label values are immutable.

### DIFF
--- a/deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
+++ b/deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
@@ -133,7 +133,7 @@ spec:
     - metadata:
         name: cache
         labels:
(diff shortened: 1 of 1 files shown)


deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
```diff
@@ -133,7 +133,7 @@ spec:
     - metadata:
         name: cache
         labels:
-          {{- include "hephaestus.buildkit.labels.standard" . | nindent 10 }}
```
sonnysideup, line 136: 🙃

sonnysideup (approved):

PR #36, opened 2022-06-03
Fixes leasing issue in workerPool.Get and adds tests

sonnysideup:
- explicitly limits leasing to pods in a _Running_ state to avoid transient state errors (think Istio!)
- adds retry logic to func that queries for buildkit pod Endpoints so that we give k8s enough time to update it
- releases previously-leased pod in the event that we cannot generate an endpoint url
- adds more unit tests for `Pool.Get`

### DIFF
--- a/pkg/buildkit/worker/pool.go
+++ b/pkg/buildkit/worker/pool.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	appsv1typed "k8s.io/client-go/kubernetes/typed/apps/v1"
 	corev1typed "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/dominodatalab/hephaestus/pkg/config"
 )
@@ -159,7 +160,16 @@ func (p *workerPool) Get(ctx context.Context) (string, error) {
 
 		// when the channel is closed, we receive nil
 		if result.pod != nil {
-			return p.buildEndpointURL(ctx, result.pod.Name)
+			addr, err := p.buildEndpointURL(ctx, result.pod.Name)
+			if err == nil {
+				return addr, nil
+			}
+			// when the url cannot be built, we release the pod
+			if rErr := p.releasePod(ctx, result.pod); rErr != nil {
+				p.log.Error(err, "Failed to release pod after URL build error")
+			}
+
+			return "", err
 		}
 	case <-ctx.Done():
 		// context has been cancelled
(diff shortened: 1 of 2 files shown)


sonnysideup:
@steved I still need to fix one test but the logic changes for the worker pool is all in place.

pkg/buildkit/worker/pool.go
```diff
@@ -240,17 +233,55 @@ func (p *workerPool) leasePod(ctx context.Context, pod *corev1.Pod) (*corev1.Pod
...
-	if err != nil {
-		return "", err
-	}
 
-	hostname, err := p.extractHostname(endpoints, podName)
+	var hostname string
+
+	// sometimes it takes a short period of time for the endpoints record to be updated with the latest list of "ready"
+	// pods, so we may need to retry fetching the resource when kubernetes experiences some lag
+	err := retry.OnError(
```
steved, line 267: FWIW there's the underlying `wait` function that powers this too - 

https://github.com/dominodatalab/hephaestus/blob/5ad22086fb05a3bafa2a552a4279098cf91d9449/pkg/buildkit/archive/archive.go#L74

pkg/buildkit/worker/pool.go
```diff
@@ -159,7 +160,16 @@ func (p *workerPool) Get(ctx context.Context) (string, error) {
 
 		// when the channel is closed, we receive nil
 		if result.pod != nil {
-			return p.buildEndpointURL(ctx, result.pod.Name)
+			addr, err := p.buildEndpointURL(ctx, result.pod.Name)
```
steved, line 163: maybe in a vacuum, `result` would have an `addr` instead of a pod and then the endpoint creation would happen during lease?

steved (approved):
No blocking comments

pkg/buildkit/worker/pool.go
sonnysideup: I saw that and considered using it until I dug into the implementation. Unfortunately, the underlying function will 
- exit the first time I return an err (i.e. i have to discard the errors in order to keep retrying)
- the final error returned after timeout is `ErrWaitTimeout`

`retry.OnError` is a bit more complicated than what I need (i.e. retriable is always true), but it does what i want otherwise (i.e. re-run on error and return the real error after timeout).

pkg/buildkit/worker/pool.go
sonnysideup: I know there's reason why I passed around a `pod` instead of an `addr` in the past but that may no longer be necessary. I think this might be a relic from the beginnings of #30 but I'll need to check just to make sure. Looking at the code, passing the address seems viable which means we could do the lease and url building inside the reconciliation loop.

ddl-dclegg (approved):

pkg/buildkit/worker/pool.go
sonnysideup: I've added this to https://github.com/dominodatalab/hephaestus/pull/37 since I had to modify pod request/result passing there anyway.

PR #37, opened 2022-06-06
Adds the ability to track "lease owner" in pod metadata

sonnysideup:
- Adds `owner` param to `Pool.Get` interface
- Adds `PodRequest` type to carry owner and results channel
- Adds `leased-at` timestamp during lease action
- `PodRequestResult` now carries an address instead of a pod
- Stores owner name in pod's lease metadata
- Now leases/builds endpoint/releases in the event of a failures in sequence w/o channels

### DIFF
--- a/pkg/buildkit/worker/pool.go
+++ b/pkg/buildkit/worker/pool.go
@@ -43,6 +43,7 @@ var (
 
 const (
 	fieldManagerName     = "hephaestus-pod-lease-manager"
+	leasedAtAnnotation   = "hephaestus.dominodatalab.com/leased-at"
 	leasedByAnnotation   = "hephaestus.dominodatalab.com/leased-by"
 	managerIDAnnotation  = "hephaestus.dominodatalab.com/manager-identity"
 	expiryTimeAnnotation = "hephaestus.dominodatalab.com/expiry-time"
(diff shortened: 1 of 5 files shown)


pkg/buildkit/worker/pool.go
```diff
@@ -441,6 +425,38 @@ func (p *workerPool) updateWorkers(ctx context.Context) error {
...
+		log.Error(err, "Failed to lease pod")
+
+		req.result <- PodRequestResult{err: err}
+		return
+	}
+
+	log.Info("Building endpoint URL")
+	addr, err := p.buildEndpointURL(ctx, pod.Name)
+
+	if err != nil {
```
steved, line 443: How common would this type of race condition (pod running, but no endpoint) happen? I'd hate to fail the build for something relatively common

steved (approved):

pkg/buildkit/worker/pool.go
sonnysideup: This is not common but it did happen to Carlin a couple of times back when there was NOT a retry strategy in place. If we `GET` the record a few milliseconds before k8s can update it with the new, ready pod's address, then this would fail. The retry should handle this w/o issue but this is just a _defensive_ guard to ensure we don't leave pods leased indefinitely.

Make sense?

steved: 🤦  forgot about the retry. Makes sense!

PR #42, opened 2022-06-08
Uses EndpointSlice watch to deliver buildkit hostname

sonnysideup:
We left ourselves open to a number of synchronization issues related to buidlkit pod addresses when we removed the "ready replicas" statefulset check in #30. To list a couple of these:

- A running pod is not guaranteed to have a registered endpoint
- A registered endpoint does not guarantee readiness

The `Endpoints` API appears to be one that the k8s maintainers look upon with contempt because of its lack of scalability, and they promote the `EndpointSlices` API as a better alternative. It turns out that the latter has a much better endpoint structure (i.e. no nested lists) and exposes a `Conditions` field.

Using a watch against one or more endpoint slices appears to resolve the issues we were seeing before while testing.

### DIFF
--- a/pkg/buildkit/worker/pool.go
+++ b/pkg/buildkit/worker/pool.go
@@ -267,14 +267,13 @@ func (p *workerPool) buildEndpointURL(ctx context.Context, podName string) (stri
 	}
 	defer watcher.Stop()
 
-	var handled bool
 	var hostname string
 
 	start := time.Now()
 	for event := range watcher.ResultChan() {
 		endpointSlice := event.Object.(*discoveryv1.EndpointSlice)
 
-		if handled, hostname = p.extractHostname(endpointSlice, podName); handled {
+		if hostname = p.extractHostname(endpointSlice, podName); hostname != "" {
 			break
 		}
 	}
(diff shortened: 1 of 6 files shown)


sonnysideup:
@ddl-cfuerst feel free to chime in here if you see fit.

pkg/buildkit/worker/pool.go
```diff
@@ -289,31 +292,40 @@ func (p *workerPool) buildEndpointURL(ctx context.Context, podName string) (stri
...
-					if port.Port == p.servicePort {
-						hostname = strings.Join([]string{address.Hostname, endpoints.Name, endpoints.Namespace}, ".")
-						p.log.Info("Found eligible endpoint address", "hostname", hostname)
-
-						break scan
-					}
-				}
-			}
+// generates internal hostname for pod using an endpoint slice
+func (p *workerPool) extractHostname(epSlice *discoveryv1.EndpointSlice, podName string) (handled bool, hostname string) {
```
steved, line 296: I don't totally understand why `hostname` can't just be checked against `""` instead of the additional `handled`?

steved (approved):

ddl-cfuerst:
No feedback except to say it works beautifully 🔥

pkg/buildkit/worker/pool.go
sonnysideup: I mean, you're totally right, I just thought a boolean felt more definitive since it's a binary "yes" or "no". I guess in this case no is `""` and anything else is yes. Truthy strings from Ruby surface yet again! 😛

PR #43, opened 2022-06-17
Reorders helm chart values

sonnysideup:
Mesting vector and manager values under "controller" and keeping shared values at the top level

### DIFF
--- a/deployments/helm/hephaestus/values.yaml
+++ b/deployments/helm/hephaestus/values.yaml
@@ -32,124 +32,11 @@ istio:
   # network settings when CNI plugin is NOT installed.
   cni: false
 
-# Controller manager configuration
+# Controller configuration
 controller:
-  # Controller manager image
-  # The default image tag is the chart appVersion.
-  image:
-    registry: ghcr.io
-    repository: dominodatalab/hephaestus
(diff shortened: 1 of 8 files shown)


deployments/helm/hephaestus/values.yaml
```diff
@@ -114,88 +59,140 @@ controller:
...
-    registry: ""
-    repository: timberio/vector
-    tag: 0.21.X-distroless-static
-    pullPolicy: IfNotPresent
-    pullSecrets: []
-
-  # Enable the deployment of the log processing sidecar
-  enabled: true
+  # Additional volume mounts added to both the manager and vector containers
+  extraVolumeMounts: []
```
steved, line 87: Feels a little "safer" to have this split into manager and vector extraVolumeMounts, but I also buy YAGNI

steved (approved):

deployments/helm/hephaestus/values.yaml
sonnysideup: We really only need this for right now but yeah I thought about that, and I think the middle ground is probably to have both a _unified_ and _separate_ `extraVolumeMounts`.

PR #45, opened 2022-06-23
Adds ListWorkers check when building buildkit client

sonnysideup:
a new buildkit client does not verify its connection with a buildkitd
worker when first created so a garbage addr like "foo.bar:1234" will
not return an error. running a "list workers" operation like we do in
the statefulset readiness check for buildkit should solve this problem
and also guarantee that the TCP server has had time to set up its TLS
certificates.

retrying the operation with a backoff makes several attempts to
establish the connection before the client is returned to the caller

### DIFF
--- a/pkg/buildkit/buildkit.go
+++ b/pkg/buildkit/buildkit.go
@@ -73,7 +73,7 @@ func (b *clientBuilder) Build(ctx context.Context) (*Client, error) {
 	b.log.Info("Confirming buildkitd connectivity")
 	err = wait.ExponentialBackoffWithContext(ctx, clientCheckBackoff, func() (done bool, err error) {
 		if _, lastErr = bk.ListWorkers(ctx); lastErr != nil {
-			b.log.Info("Buildkitd is not ready")
+			b.log.V(1).Info("Buildkitd is not ready")
 			return false, nil
 		}
 
@@ -82,6 +82,7 @@ func (b *clientBuilder) Build(ctx context.Context) (*Client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to contact buildkitd after %d attempts: %w", clientCheckBackoff.Steps, lastErr)
 	}
+	b.log.Info("Buildkitd connectivity established")
 
 	return &Client{
 		bk:               bk,
--- a/pkg/buildkit/buildkit.go
+++ b/pkg/buildkit/buildkit.go
@@ -23,10 +23,10 @@ import (
 	"github.com/dominodatalab/hephaestus/pkg/buildkit/archive"
 )
 
-var clientCheckBackoff = wait.Backoff{ // retries after 10ms 50ms 250ms 1250ms 6250ms
+var clientCheckBackoff = wait.Backoff{ // retries after 500ms 1s 2s 4s 8s with jitter
 	Steps:    5,
-	Duration: 10 * time.Millisecond,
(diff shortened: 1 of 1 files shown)


pkg/buildkit/buildkit.go
```diff
@@ -18,10 +18,18 @@ import (
 	"github.com/moby/buildkit/util/progress/progressui"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/dominodatalab/hephaestus/pkg/buildkit/archive"
 )
 
+var clientCheckBackoff = wait.Backoff{ // retries after 10ms 50ms 250ms 1250ms 6250ms
```
fraenkel, line 26: seems quite fast relative to the time it might take.
does it really mater if its 10ms, 50ms 250ms, vs the more normal 1s, 2s , 4s...

pkg/buildkit/buildkit.go
```diff
@@ -60,6 +68,22 @@ func (b *clientBuilder) Build(ctx context.Context) (*Client, error) {
...
+	err = wait.ExponentialBackoffWithContext(ctx, clientCheckBackoff, func() (done bool, err error) {
+		if _, lastErr = bk.ListWorkers(ctx); lastErr != nil {
+			b.log.V(1).Info("Buildkitd is not ready")
+			return false, nil
+		}
+
+		return true, nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to contact buildkitd after %d attempts: %w", clientCheckBackoff.Steps, lastErr)
```
fraenkel, line 83: shouldn't you just log the failure here and return the error instead?

fraenkel (approved):

pkg/buildkit/buildkit.go
sonnysideup: We running a test we found that under light load that it took < 1s but > 250ms. I think I'm just trying to be as fast as possible here but you're probably right. Maybe something like 500ms, 1s, 2s, 4s... seems more reasonable?

pkg/buildkit/buildkit.go
sonnysideup: I wish. `wait.ExponentialBackoffWithContext` will stop immediately if you return an error. 😞 

https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoff

pkg/buildkit/buildkit.go
fraenkel: good compromise

pkg/buildkit/buildkit.go
fraenkel: it does

https://github.com/kubernetes/apimachinery/blob/v0.24.2/pkg/util/wait/wait.go#L421

sonnysideup: I guess in this case I don't care about the err, maybe that's wrong. I just want the operation to succeed.

fraenkel: Correct, but normally I would let the pod crashloop so alerting can take over as it prolongs. So its a balance between retrying enough times vs giving up.
If o11y was a thing, crashlooping is way easier to monitor than adding more metrics to report a prolonged issue where no progress is made.

PR #50, opened 2022-07-13
support configuration of insecure registries

steved:
In Forge, we allowed setting of insecure registries [on the objects themselves](https://github.com/dominodatalab/hephaestus/blob/747da51794616fe700c302e7c21c18a842e8159e/pkg/api/forge/v1alpha1/containerimagebuild_types.go#L64), which did not make it to Hephaestus.

Unfortunately, the only way to configure HTTP-only or insecure HTTPS registry hosts is [through the buildkitd.toml](https://github.com/moby/buildkit/issues/2044#issuecomment-811489932) config. This makes it impossible to set them per-image build and requires deploy-time configuration.

To support credential validation against HTTP-only or insecure registries, the info needs to be passed to the controller as well.

### DIFF
--- a/deployments/helm/hephaestus/templates/buildkit/configmap.yaml
+++ b/deployments/helm/hephaestus/templates/buildkit/configmap.yaml
@@ -27,4 +27,15 @@ data:
         all = true
         keepBytes = {{ .gcKeepStorage | int64 }}
       {{- end }}
+
+    {{- range $domain, $opts := $.Values.registries }}
+    [registry."{{ $domain }}"]
+      {{- with $opts.http }}
+      http = {{ . }}
+      {{- end }}
+
+      {{- with $opts.insecure }}
+      insecure = {{ . }}
+      {{- end }}
+    {{- end }}
   {{- end }}
--- a/deployments/helm/hephaestus/values.yaml
+++ b/deployments/helm/hephaestus/values.yaml
@@ -43,6 +43,13 @@ newRelic:
   # Tag metadata added to metrics
   labels: {}
 
+# Configuration for buildkit and controller that adds the ability to pull/push images
+# from/to insecure (self-signed TLS) and http registries.
+registries: {}
+  # myserver:
+  #   insecure: true
+  #   http: true
+
 # Controller configuration
 controller:
   # Number of instances to run. Leader election will be enabled whenever this
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -93,11 +93,18 @@ type Buildkit struct {
 
 	Secrets map[string]string `json:"secrets" yaml:"secrets,omitempty"`
 
+	Registries map[string]RegistryConfig `json:"registries,omitempty" yaml:"registries,omitempty"`
+
 	PoolSyncWaitTime *time.Duration `json:"poolSyncWaitTime" yaml:"poolSyncWaitTime"`
 	PoolMaxIdleTime  *time.Duration `json:"poolMaxIdleTime" yaml:"poolMaxIdleTime"`
 	PoolWatchTimeout *int64         `json:"poolWatchTimeout" yaml:"poolWatchTimeout"`
 
 	MTLS *BuildkitMTLS `json:"mtls,omitempty" yaml:"mtls,omitempty"`
+}
+
+type RegistryConfig struct {
+	Insecure bool `json:"insecure,omitempty" yaml:"insecure,omitempty"`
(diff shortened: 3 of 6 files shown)


deployments/helm/hephaestus/templates/buildkit/configmap.yaml
```diff
@@ -27,4 +27,15 @@ data:
...
+    {{- range $server, $opts := $.Values.registries }}
+    [registry."{{ $server }}"]
+      {{- with $opts.http }}
+      http = {{ . }}
+      {{- end }}
+
+      {{- with $opts.insecure }}
+      insecure = {{ . }}
+      {{- end }}
+    {{- end }}
```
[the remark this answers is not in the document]
steved: Unfortunately `registry.insecure=true` only impacts the registry being pushed to and not the registry being pulled from (which we also need).

pkg/config/config.go
```diff
@@ -93,13 +93,20 @@ type Buildkit struct {
 
 	Secrets map[string]string `json:"secrets" yaml:"secrets,omitempty"`
 
+	Registries map[string]*RegistryConfig `json:"registries,omitempty" yaml:"registries,omitempty"`
```
sonnysideup, line 96: Should this be a pointer? We wouldn't want a a mapping to nil and we don't need to modify it, right?

pkg/controller/imagebuild/component/builddispatcher.go
```diff
@@ -104,7 +104,15 @@ func (c *BuildDispatcherComponent) Reconcile(ctx *core.Context) (ctrl.Result, er
...
 
 	validateCredsSeg := txn.StartSegment("credentials-validate")
-	if err = credentials.Verify(ctx, configDir); err != nil {
+
+	insecureRegistries := make([]string, 0)
+	for reg, opts := range c.cfg.Registries {
+		if opts.Insecure || opts.HTTP {
+			insecureRegistries = append(insecureRegistries, reg)
+		}
+	}
```
sonnysideup, line 113: There's a part of me that wants to use `Config.Validate` to disallow registry entries with both insecure/http fields false so we wouldn't have to perform this check here, but I think your design is more flexible and still works. Nothing to change here, just thinking via comments.

deployments/helm/hephaestus/values.yaml
```diff
@@ -43,6 +43,12 @@ newRelic:
   # Tag metadata added to metrics
   labels: {}
 
+# Buildkit registry configuration
```
sonnysideup, line 46: I might add a little more info to this comment.

```suggestion
# Configuration for buildkit and controller that adds the ability to pull/push images
# from/to insecure (self-signed TLS) and http registries.
```

deployments/helm/hephaestus/templates/buildkit/configmap.yaml
sonnysideup, line 40: I know this will configure "pull" from insecure/http, but does it also handle "push", or do we also need to pass that information to the buildkit client?

steved: Confirmed it works for both by setting the internal DR to `http` and got back a 400.

pkg/config/config.go
steved: Nope, just instinct. Updated.

pkg/controller/imagebuild/component/builddispatcher.go
```diff
@@ -104,24 +107,39 @@ func (c *BuildDispatcherComponent) Reconcile(ctx *core.Context) (ctrl.Result, er
...
+
+	buildLog.Info("Validating registry credentials")
+	if err = credentials.Verify(ctx, configDir, insecureRegistries); err != nil {
 		txn.NoticeError(newrelic.Error{
 			Message: err.Error(),
 			Class:   "CredentialsValidateError",
 		})
+
+		buildLog.Error(err, fmt.Sprintf("Failed to validate registry credentials: %s", err.Error()))
```
steved, line 125: This prints an error like this when the registry is plainhttp without a valid config:
```
Jul 14 2022 18:37:13 -0700 | Validating registry credentials
Jul 14 2022 18:37:13 -0700 | Failed to validate registry credentials: "10.0.0.8:5001" client credentials are invalid: Get "https://10.0.0.8:5001/v2/": http: server gave HTTP response to HTTPS client  
```

sonnysideup (approved):

PR #52, opened 2022-07-19
Improves Java SDK generation

sonnysideup:
- Fixed time serialization bug that caused the `list/get` operations to fail
- Fixed token-based authentication by adding _bearer token_ auth strategy to client
- Pushes artifacts to artifactory for consumption by application engineers

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -255,14 +255,14 @@ jobs:
       - name: Publish JAR to GitHub
         run: |
           docker run --rm \
-          --workdir /wd \
-          --volume $HOME/.m2:/root/.m2 \
-          --volume $(pwd)/sdks/java:/wd \
-          --env ARTIFACTORY_USERNAME=${{ secrets.ARTIFACTORY_USERNAME }} \
-          --env ARTIFACTORY_PASSWORD=${{ secrets.ARTIFACTORY_PASSWORD }} \
-          --env GITHUB_USERNAME=${{ github.actor }} \
-          --env GITHUB_PASSWORD=${{ secrets.GITHUB_TOKEN }} \
-          $MAVEN_DOCKER_IMAGE mvn --settings settings.xml --activate-profiles github -DskipTests deploy
+            --workdir /wd \
+            --volume $HOME/.m2:/root/.m2 \
+            --volume $(pwd)/sdks/java:/wd \
+            --env ARTIFACTORY_USERNAME=${{ secrets.ARTIFACTORY_USERNAME }} \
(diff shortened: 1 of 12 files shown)


.github/workflows/main.yml
```diff
@@ -207,22 +229,49 @@ jobs:
...
           path: sdks/java/target/*.jar
           if-no-files-found: error
 
-      - name: Publish Java JAR
-        run: docker run --rm -v $HOME/.m2:/root/.m2 -v $(pwd)/sdks/java:/wd -e JAVA_SDK_MAVEN_USERNAME=${{ github.actor }} -e JAVA_SDK_MAVEN_PASSWORD=${{ secrets.GITHUB_TOKEN }} --workdir /wd maven:3-openjdk-8 mvn -s settings.xml deploy -DskipTests -DaltDeploymentRepository=github::default::https://maven.pkg.github.com/dominodatalab/hephaestus
-        if: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v') }}
+      - name: Publish JAR to GitHub
+        run: |
+          docker run --rm \
+          --workdir /wd \
```
steved, line 258: nit, but indent the lines after the first

build/README.md
```diff
@@ -5,5 +5,6 @@ the GitHub workflow, and pushed to the appropriate location.
 
 ## Buildkit
 
-The rootless image for buildkit has been modified to expand the uid/gid map
-range and accommodate environments where Istio is running.
+All buildkit images have had their APKdependencies upgraded, and the rootless
```
steved, line 8:
```suggestion
All buildkit images have had their APK dependencies upgraded, and the rootless
```

prototype/openapiv3/main.go
```diff
@@ -0,0 +1,219 @@
...
+// 	}
+//
+// 	for _, response := range op.Responses.StatusCodeResponses {
+// 		for _, mediaType := range response.Content {
+// 			if schema := mediaType.Schema; schema != nil {
+// 				schema.Ref = refFunc(schema.Ref.String())
+// 			}
+// 		}
+// 	}
+// }
```
steved, line 202: Is this going to be used for something in a follow-up?

steved (approved):

prototype/openapiv3/main.go
sonnysideup: Yeah, the `/openapi/v3` endpoint is much more user-friendly than the v2 one because it returns targeted resources (i.e. `/openapi/v3/apis/hephaestus.dominodatalab.com/v1`) instead of EVERYTHING. It's useful but not critical for this work so I'm going to do that work slowly and deliberately as a separate PR when I have time. These are just my notes for that future work.

PR #53, opened 2022-08-04
Exposes the endpointSliceWatchTimeout value

sonnysideup:
@steved I found this [interesting article](https://engineering.dollarshaveclub.com/kubernetes-fixing-delayed-service-endpoint-updates-fd4d0a31852c) that describes some of the issues we see in environments where buildkit pods are frequently terminated and re-created.

My conclusion is that (a) my static 90s endpoint watch timeout is too low in certain instances where the buildkit pods churn a lot and (b) this value should be externally configurable so that it can serviced without code changes.

## Changes

- Exposes `fetchAndExtractTimeout` value to control archive processing time and bumps default fetch retry considerably
- Exposes `poolEndpointWatchTimeout` to allow the configuration of this value
- Upgrades to go 1.19
- Upgrades vector to 0.23
- Upgrades `golangci-lint`
- Adds diagnostic logging whenever we fail to acquire a buildkit pod endpoint


pkg/buildkit/worker/pool.go
```diff
@@ -466,8 +466,8 @@ func (p *workerPool) processPodRequest(ctx context.Context, req *PodRequest, pod
 	if err != nil {
 		log.Error(err, "Failed to build routable URL")
 
-		if rErr := p.releasePod(ctx, pod); rErr != nil {
-			log.Error(err, "Failed to release pod")
+		if rErr := p.releasePod(ctx, &pod); rErr != nil {
+			log.Error(rErr, "Failed to release pod")
```
sonnysideup, line 470: This was logging the wrong error.

pkg/buildkit/worker/pool.go
```diff
@@ -375,8 +376,7 @@ func (p *workerPool) updateWorkers(ctx context.Context) error {
 			if req := p.requests.Dequeue(); req != nil {
 				log.Info("Found pending pod request, processing")
 
-				pod := pod
-				if p.processPodRequest(ctx, req, &pod) {
+				if p.processPodRequest(ctx, req, pod) {
```
sonnysideup, line 379: Just passing the pod value instead of a pointer to avoid implicit memory aliasing and variable re-declaration.

steved (approved):

PR #55, opened 2022-08-17
Fix worker pool

sonnysideup:
See the commit extended comments

### DIFF
--- a/pkg/buildkit/worker/pool.go
+++ b/pkg/buildkit/worker/pool.go
@@ -51,7 +51,7 @@ const (
 	expiryTimeAnnotation = "hephaestus.dominodatalab.com/expiry-time"
 )
 
-type workerPool struct {
+type AutoscalingPool struct {
 	log logr.Logger
 
 	// system shutdown
@@ -89,7 +89,12 @@ type workerPool struct {
 }
 
 // NewPool creates a new worker pool that can be used to lease buildkit workers for image builds.
-func NewPool(ctx context.Context, clientset kubernetes.Interface, conf config.Buildkit, opts ...PoolOption) Pool {
+func NewPool(
+	ctx context.Context,
+	clientset kubernetes.Interface,
+	conf config.Buildkit,
+	opts ...PoolOption,
+) *AutoscalingPool {
 	o := defaultOpts
 	for _, fn := range opts {
 		o = fn(o)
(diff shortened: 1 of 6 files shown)


pkg/controller/start.go
```diff
@@ -42,7 +42,7 @@ func Start(cfg config.Controller) error {
 	ctrl.SetLogger(zapr.NewLogger(zapLogger))
 
 	log := ctrl.Log.WithName("setup")
-	log.V(1).Info("Using provided configuration", "config", cfg)
+	log.Info("Using provided configuration", "config", cfg)
```
steved, line 45: This doesn't have secrets in it?

pkg/buildkit/worker/pool.go
```diff
@@ -502,8 +520,44 @@ func (p *workerPool) triggerReconcile() {
 	}
 }
 
+// ensure pod is operational by checking its phase and conditions
+func (p *AutoscalingPool) isOperationalPod(ctx context.Context, podName string) (verdict bool) {
+	// fetch the latest version of the pod
+	pod, err := p.podClient.Get(ctx, podName, metav1.GetOptions{})
```
steved, line 526: pretty minor, but should this have a resource version associated with the request (i.e. no older than the one we got from the list)?

steved (approved):

pkg/controller/start.go
sonnysideup: mmm, the amqp url 😬 . i'll have to figure out how logr serializes the object and make sure that field and any others get sanitized. seems like a worthwhile effort.

pkg/buildkit/worker/pool.go
sonnysideup: maybe? i look at this again, thanks for raising it.

sonnysideup: @steved https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list

> For **get**, the semantics of resourceVersion are: `resourceVersion` unset == "Most Recent"

steved: Makes sense, I was just curious if we should be explicit. Thanks for checking.

pkg/controller/start.go
sonnysideup: We can just implement `MarshalJSON` for any of the configuration structs whenever we want to _sanitize_ rendered data.

PR #57, opened 2022-08-24
Removes server name checking from secret-based registry auth

sonnysideup:
If a user wants to authenticate with 2 or more registries contained within a single k8s secret, then they are required to create an `ImageBuild` with the following auth section:

```yaml
registryAuths:
- server: "server1"
  secret:
    name: many-registry-auths
    namespace: ns
- server: "server2"
  secret:
    name: many-registry-auths
    namespace: ns
```

and this can become somewhat cumbersome as that list grows. Furthermore, the `server` field for secrets is meant to server as a verification mechanism (i.e. an error will be raised if it's missing from the underlying secret data). I think we can add the verification to another field, or implement a different way that removes the labor involved here. For now, I'm disabling it and promoting all auths contained within the secret over to buildkit.

/cc @ddl-kevin

### DIFF
--- a/pkg/api/hephaestus/v1/types.go
+++ b/pkg/api/hephaestus/v1/types.go
@@ -37,6 +37,11 @@ type SecretCredentials struct {
 }
 
 type RegistryCredentials struct {
+	// NOTE: this field was previously used to assert the presence of an auth entry inside of secret credentials. if the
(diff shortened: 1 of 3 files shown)


ddl-kevin (approved):
LGTM

saahildhulla (approved):
👍

PR #60, opened 2022-09-02
DOM-38748: Harden security contexts

fraenkel:


deployments/helm/hephaestus/templates/controller/deployment.yaml
```diff
@@ -117,6 +117,10 @@ spec:
         {{- if .Values.controller.vector.enabled }}
         - name: vector
           securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+                drop:
```
steved, line 122: indent is off?

steved (approved):

PR #61, opened 2022-09-02
PLAT-5472 update to chart with default values

gerardo8a:
Add pod default label, annotations and env values to pods. This is part of the standardization effort.
https://dominodatalab.atlassian.net/wiki/spaces/ENG/pages/2298773600/Helm+Charts+Standardization

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ bin/
 
 # Generated files for local development
 local-development/
+
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@ bin/
 
 # Generated files for local development
 local-development/
-
-.idea/
--- a/deployments/helm/hephaestus/Chart.yaml
+++ b/deployments/helm/hephaestus/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: hephaestus
-version: 0.0.1
+version: 0.0.0
 appVersion: "latest"
 kubeVersion: ">= 1.19.0-0"
 description: Kubernetes operator that builds OCI images using buildkit.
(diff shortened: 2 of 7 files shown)


deployments/helm/hephaestus/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: hephaestus
-version: 0.0.0
+version: 0.0.1
```
steved, line 4: This isn't necessary, it's auto-filled in at release time to the correct tag

steved (approved):

deployments/helm/hephaestus/Chart.yaml
gerardo8a: got it, change reverted

deployments/helm/hephaestus/values.yaml
```diff
@@ -24,6 +24,15 @@ enablePodSecurityPolicies: false
...
 
+# Annotations applied to pods
+podAnnotations: {}
+
+# Labels applied to pods
+podLabels: {}
+
+# Environment applied to pods
+podEnv: {}
+
```
sonnysideup, line 35: Maybe just indicate `applied to all pods`? It's implicit since it's a top-level config but I always try to error on the side of being explicit.

.gitignore
```diff
@@ -9,3 +9,5 @@ bin/
 
 # Generated files for local development
 local-development/
+
+.idea/
```
sonnysideup, line 13: In general, I like to leave IDE-specific ignore statements out of my project and have devs add them to their global ignore file (https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) but whatevs.

sonnysideup (approved):

.gitignore
gerardo8a: I can do that no worries, I don't have a problem with that

PR #66, opened 2022-09-21
Moves from discoveryv1beta1 to discoveryv1 API for EndpointSlices

sonnysideup:
The former has been removed in k8s 1.25 since we no longer need to support k8s < 1.21. _Mo' versions, mo' problems._


JadeHayes (approved):
Lgtm!

PR #77, opened 2022-11-02
PLAT-5844: Allow buildkit groups to runAsAny

fraenkel:
Given PSPs are going away shortly, simplify the PSP to work in root and rootless mode.
Currently when running as root, the PSP forces the fsGroup is set to 1. Once PSPs go away, the fsGroup will switch to 0.

The current securityContext is setting the fsGroup to match the user id and failing. Rather than play games with the PSP temporarily just allow it to runAsAny which will allow both settings to work.


miguelhar (approved):

sonnysideup (approved):
Seems like a good enough stopgap

PR #82, opened 2022-11-30
ensure that leased, pending, or running pods are never scaled down during pool reconciliation

steved:
27c43734053bc3095972a0c83a5b1d2040bca0de contains the failing test case: a leased pod was removed when only the last expired pod should have been.

193bd27674903861e228c4bbfc484b4fe4c6648b contains a rework of the pool arbitration to ensure that no leased, pending, or running pods are scaled down. only above that minimum number of replicas can we prioritize recycling of expired/invalid pods over serving a new build request.


steved:
Run the functional tests please?

sonnysideup (approved):

PR #83, opened 2022-12-05
Replace AMQP library

sonnysideup:
Using external library now that does not have race conditions causing a deadlock when closing the client during a a _reconnect event_.


sonnysideup:
resolves #83

test/functional/helpers_test.go
```diff
@@ -632,7 +633,9 @@ func testMessageDelivery(t *testing.T, ctx context.Context, client kubernetes.In
 		hostname = svc.Status.LoadBalancer.Ingress[0].IP
 	}
 	rmqURL := fmt.Sprintf("amqp://user:rabbitmq-password@%s:5672/", hostname)
-	conn, channel, err := amqp.Dial(rmqURL)
+	conn, err := amqp091.Dial(rmqURL)
+	require.NoError(t, err, "failed to connet to rabbitmq service")
```
steved, line 637:
```suggestion
	require.NoError(t, err, "failed to connect to rabbitmq service")
```

pkg/messaging/amqp/publisher.go
```diff
@@ -1,178 +0,0 @@
-package amqp
```
steved, line 1: Do you want to remove `manager.go` too?

steved (approved):

PR #86, opened 2023-03-21
* Adding ib cleanup polling.

JadeHayes:


pkg/config/config.go
```diff
@@ -83,6 +83,8 @@ type Manager struct {
 	WebhookPort          int      `json:"webhookPort" yaml:"webhookPort"`
 	WatchNamespaces      []string `json:"watchNamespaces" yaml:"watchNamespaces,omitempty"`
 	EnableLeaderElection bool     `json:"enableLeaderElection" yaml:"enableLeaderElection"`
+	MaxIBRetentionCount  int      `json:"maxIBRetentionCount" yaml:"maxIBRetentionCount"`
```
JadeHayes, line 86: I'm not sure where to set this for production Hephaestus. (Where is the hephaestus.yaml when deploying Hephaestus)

PR #87, opened 2023-03-23
* Added IB Clean up.

JadeHayes:
### Link to JIRA

[PLAT-5024](https://dominodatalab.atlassian.net/browse/PLAT-5024)

### What issue does this pull request solve?

* IBs in a "finished" state are not cleaned up, these IBs start to pile up and eat up resources. 

### What is the solution?

* Cronjob to clean up periodically, default daily.

### Testing
_Ran Hephaestus locally and created cronjobs that removed additional IBs in a "finished" state_

- [x] Unit test(s)

### References (optional)

_N/A_

### DIFF
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -49,10 +49,18 @@ spec:
                 - run-gc
                 - --maxIBRetention={{ .Values.controller.ibGarbageCollector.maxIBRetention | int }}
                 - --config=/etc/hephaestus/config.yaml
-              {{- with .Values.controller.manager.extraEnvVars  }}
+              {{- with .Values.controller.manager }}
+              {{- if or .extraEnvVars $.Values.podEnv }}
               env:
-                {{- include "common.tplvalues.render" (dict "value" . "context" $) | nindent 12 }}
+              {{- with .extraEnvVars }}
+                {{- include "common.tplvalues.render" (dict "value" . "context" $) | nindent 14 }}
               {{- end }}
+              {{- with $.Values.podEnv }}
+                {{- toYaml . | nindent 14 }}
+              {{- end }}
+              {{- end }}
+              {{- end }}
+
           {{- with .Values.controller.nodeSelector }}
           nodeSelector:
             {{- toYaml . | nindent 12 }}
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -12,6 +12,7 @@ spec:
             - name: config-vol
               secret:
                 secretName: {{ include "hephaestus.configSecretName" . }}
+          serviceAccountName: {{ include "hephaestus.serviceAccountName" . }}
           containers:
             - name: image-build-gc
               image: {{ include "hephaestus.manager.image" . }}
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -1,7 +1,7 @@
 apiVersion: batch/v1
 kind: CronJob
 metadata:
-  name: {{ include "hephaestus.manager.name" . }}
+  name: {{ include "hephaestus.manager.name" . }}-cronjob
 spec:
   schedule: {{ .Values.controller.ibGarbageCollector.interval | quote }}
   jobTemplate:
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -28,6 +28,7 @@ spec:
         spec:
           securityContext:
             {{- toYaml .Values.controller.podSecurityContext | nindent 12 }}
+          restartPolicy: OnFailure
           volumes:
             - name: config-vol
               secret:
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -2,6 +2,7 @@ apiVersion: batch/v1
 kind: CronJob
 metadata:
   name: {{ include "hephaestus.manager.name" . }}-cronjob
+  labels: {{- include "hephaestus.controller.labels.standard" . | nindent 4 }}
 spec:
   schedule: {{ .Values.controller.ibGarbageCollector.interval | quote }}
   jobTemplate:
@@ -10,10 +11,6 @@ spec:
         metadata:
           annotations:
             checksum/config: {{ include (print $.Template.BasePath "/controller/secret.yaml") . | sha256sum }}
-            {{- if .Values.controller.metrics.enabled }}
-            prometheus.io/port: "{{ .Values.controller.manager.metricsPort }}"
-            prometheus.io/scrape: "true"
-            {{- end }}
             {{- with .Values.controller.podAnnotations }}
             {{- toYaml . | trimSuffix "\n" | nindent 8 }}
             {{- end }}
@@ -21,7 +18,7 @@ spec:
             {{- toYaml . | nindent 8 }}
             {{- end }}
           labels:
-            {{- include "hephaestus.controller.labels.matchLabels" . | nindent 12 }}
+              {{- include "hephaestus.controller.labels.matchLabels" . | nindent 12 }}
               {{- with .Values.controller.podLabels }}
               {{- toYaml . | trimSuffix "\n" | nindent 12 }}
               {{- end }}
@@ -50,7 +47,7 @@ spec:
               command: [ "hephaestus-controller" ]
               args:
                 - run-gc
-                - --maxIBRetention=2
+                - --maxIBRetention={{ .Values.controller.ibGarbageCollector.maxIBRetention | int }}
                 - --config=/etc/hephaestus/config.yaml
               {{- with .Values.controller.manager.extraEnvVars  }}
               env:
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -42,7 +42,7 @@ spec:
               imagePullPolicy: {{ .Values.controller.manager.image.pullPolicy }}
               volumeMounts:
                 - name: config-vol
-                  mountPath: /config.yaml
+                  mountPath: /etc/hephaestus/config.yaml
                   subPath: config.yaml
               command: [ "hephaestus-controller" ]
               args:
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -4,9 +4,10 @@ metadata:
   name: {{ include "hephaestus.manager.name" . }}-cronjob
   labels: {{- include "hephaestus.controller.labels.standard" . | nindent 4 }}
 spec:
-  schedule: {{ .Values.controller.ibGarbageCollector.interval | quote }}
+  schedule: {{ .Values.controller.completedImageBuildCleanup.interval | quote }}
   jobTemplate:
     spec:
+      backoffLimit: {{ .Values.controller.completedImageBuildCleanup.backoffLimit | int }}
       template:
         metadata:
           annotations:
@@ -18,17 +19,17 @@ spec:
             {{- toYaml . | nindent 12 }}
             {{- end }}
           labels:
-              {{- include "hephaestus.controller.labels.matchLabels" . | nindent 12 }}
-              {{- with .Values.controller.podLabels }}
-              {{- toYaml . | trimSuffix "\n" | nindent 12 }}
-              {{- end }}
-              {{- with .Values.podLabels }}
-              {{- toYaml . | nindent 12 }}
-              {{- end }}
+            {{- include "hephaestus.controller.labels.matchLabels" . | nindent 12 }}
+            {{- with .Values.controller.podLabels }}
+            {{- toYaml . | trimSuffix "\n" | nindent 12 }}
+            {{- end }}
+            {{- with .Values.podLabels }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
         spec:
           securityContext:
             {{- toYaml .Values.controller.podSecurityContext | nindent 12 }}
-          restartPolicy: OnFailure
+          restartPolicy: Never
           volumes:
             - name: config-vol
               secret:
@@ -47,7 +48,7 @@ spec:
               command: [ "hephaestus-controller" ]
               args:
                 - run-gc
-                - --maxIBRetention={{ .Values.controller.ibGarbageCollector.maxIBRetention | int }}
+                - --maxIBRetention={{ .Values.controller.completedImageBuildCleanup.retainMax | int }}
                 - --config=/etc/hephaestus/config.yaml
               {{- with .Values.controller.manager }}
               {{- if or .extraEnvVars $.Values.podEnv }}
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -12,10 +12,10 @@ spec:
           annotations:
             checksum/config: {{ include (print $.Template.BasePath "/controller/secret.yaml") . | sha256sum }}
             {{- with .Values.controller.podAnnotations }}
-            {{- toYaml . | trimSuffix "\n" | nindent 8 }}
+            {{- toYaml . | trimSuffix "\n" | nindent 12 }}
             {{- end }}
             {{- with .Values.podAnnotations }}
-            {{- toYaml . | nindent 8 }}
+            {{- toYaml . | nindent 12 }}
             {{- end }}
           labels:
               {{- include "hephaestus.controller.labels.matchLabels" . | nindent 12 }}
@@ -56,15 +56,15 @@ spec:
           restartPolicy: OnFailure
           {{- with .Values.controller.nodeSelector }}
           nodeSelector:
-          {{- toYaml . | nindent 8 }}
+          {{- toYaml . | nindent 10 }}
           {{- end }}
           {{- with .Values.controller.affinity }}
           affinity:
-            { { - toYaml . | nindent 8 }}
+            {{- toYaml . | nindent 10 }}
           {{- end }}
           {{- with .Values.controller.tolerations }}
           tolerations:
-            {{- toYaml . | nindent 8 }}
+            {{- toYaml . | nindent 10 }}
           {{- end }}
           {{- with .Values.controller.priorityClassName }}
           priorityClassName: {{ . | quote }}
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -52,14 +52,10 @@ spec:
                 - run-gc
                 - --maxIBRetention=2
                 - --config=/etc/hephaestus/config.yaml
-              {{- with .Values.controller.manager }}
-              {{- if .extraEnvVars  }}
+              {{- with .Values.controller.manager.extraEnvVars  }}
               env:
-                {{- with .extraEnvVars }}
                 {{- include "common.tplvalues.render" (dict "value" . "context" $) | nindent 12 }}
-                {{- end }}
-              {{ end }}
-              {{ end }}
+              {{- end }}
           restartPolicy: OnFailure
           {{- with .Values.controller.nodeSelector }}
           nodeSelector:
--- a/deployments/helm/hephaestus/templates/controller/cronjob.yaml
+++ b/deployments/helm/hephaestus/templates/controller/cronjob.yaml
@@ -50,7 +50,6 @@ spec:
               command: [ "hephaestus-controller" ]
               args:
                 - run-gc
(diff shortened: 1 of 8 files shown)


deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: image-build-gc
+spec:
+  schedule: "* * * * *"
```
JadeHayes, line 6: Need to make this configurable from the values.yaml

pkg/controller/runGC.go
```diff
@@ -0,0 +1,143 @@
...
+	return &ImageBuildGC{
+		maxIBRetention: maxIBRetention,
+		hephClient:     hephClient,
+		namespaces:     ns,
+	}, nil
+}
+
+func getAllNamespaces(config *rest.Config) ([]string, error) {
+	// create k8s client to access all namespaces in the cluster
+	k8sClient, err := k8s.NewForConfig(config)
```
JadeHayes, line 62: I was also thinking another route would be to add the k8s client to the Hephaestus client. Seemed weird to change the client for this thought

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,32 @@
...
+                  mountPath: /etc/hephaestus/config.yaml
+                  subPath: config.yaml
+              imagePullPolicy: IfNotPresent
+              command: [ "hephaestus-controller" ]
+              args:
+                - run-gc
+                - --enabled=true
+                - --maxRetention=2
+                - --config=/etc/hephaestus/config.yaml
+                - echo IB Cron job GC started
```
steved, line 31: Remove this?
```suggestion
                - echo IB Cron job GC started
```

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,32 @@
...
+                - name: config-vol
+                  readOnly: true
+                  mountPath: /etc/hephaestus/config.yaml
+                  subPath: config.yaml
+              imagePullPolicy: IfNotPresent
+              command: [ "hephaestus-controller" ]
+              args:
+                - run-gc
+                - --enabled=true
+                - --maxRetention=2
```
steved, line 29: You have `maxIBRetention` in the controller command, but `maxRetention` here

pkg/cmd/controller/root.go
```diff
@@ -18,8 +18,15 @@ func NewCommand() *cobra.Command {
 		Short: "OCI image build controller using buildkit",
 	}
 	cmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "hephaestus.yaml", "configuration file")
+
+	runGCCmd := newRunGCCommand()
+	runGCCmd.Flags().Bool("enabled", true, "Enable the auto image builder clean up.")
```
steved, line 23: Do we need an enabled flag in the CLI? Can we just enable/disable the cronjob?

pkg/cmd/controller/root.go
```diff
@@ -18,8 +18,15 @@ func NewCommand() *cobra.Command {
...
+	runGCCmd.Flags().Int("maxIBRetention", 5, "Maximum number of image builds to retain."+
```
steved, line 24: Just in the interest of being consistent, the flags should be defined in the `newXYZCommand()` method like the others?

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: image-build-gc
+spec:
+  schedule: "* * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
```
steved, line 10: We'll want to make sure we support the same configurables and defaults as the manager deployment: 
https://github.com/dominodatalab/hephaestus/blob/main/deployments/helm/hephaestus/templates/controller/deployment.yaml#L36

* pod labels, annotations
* pod security context
* container security context
* image pull policy (`imagePullPolicy: {{ .Values.controller.manager.image.pullPolicy }}`)
* `extraEnvVars`
* we should consider having some `resources` definition, even if it's quite low
* nodeSelector
* affinity
* tolerations
* priority class

For the majority of these you can use the same Helm values as the manager. Resources might be one to keep separate.

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: image-build-gc
```
steved, line 4: Should probably have a `{{ include "hephaestus.manager.name" . }}` prefix and the cronjob needs standard labels and annotations

pkg/controller/runGC.go
```diff
@@ -0,0 +1,143 @@
...
+	if len(ibNamespaces) > 0 {
+		log.Info("Limiting GC cleanup", "namespaces", ibNamespaces)
+		ns = ibNamespaces
+	} else {
+		ns, err = getAllNamespaces(config)
+		if err != nil {
+			log.Info("Unable to access cluster namespaces")
+			return nil, err
+		}
+		log.Info("Watching all namespaces")
```
steved, line 50:
```suggestion
		log.Info("Running GC against all namespaces")
```

pkg/controller/runGC.go
```diff
@@ -0,0 +1,143 @@
...
+	for _, namespace := range namespaces.Items {
+		ns = append(ns, namespace.Name)
+	}
+	return ns, nil
+}
+
+func (gc *ImageBuildGC) CleanUpIBs(ctx context.Context, log logr.Logger, namespace string) {
+	crdList, err := gc.hephClient.HephaestusV1().ImageBuilds(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		log.Info("Unable to access a list of IBs, not starting IB clean up", "error", err)
```
steved, line 81: Feels like we should bubble up actual errors and then exit 1 if any fail?

pkg/controller/runGC.go
```diff
@@ -0,0 +1,143 @@
...
+			"resourceCount", len(builds), "retentionCount", gc.maxIBRetention)
+		return
+	}
+	log.Info("Total resources eligible for deletion", "count", len(builds))
+	sort.Slice(builds, func(i, j int) bool {
+		return builds[i].CreationTimestamp.Before(&builds[j].CreationTimestamp)
+	})
+	deletePolicy := metav1.DeletePropagationForeground
+	for _, build := range builds[:len(builds)-gc.maxIBRetention] {
+		if err := gc.hephClient.HephaestusV1().ImageBuilds("default").Delete(ctx, build.Name,
```
steved, line 110:
```suggestion
		if err := gc.hephClient.HephaestusV1().ImageBuilds(namespace).Delete(ctx, build.Name,
```

pkg/controller/runGC.go
```diff
@@ -0,0 +1,143 @@
...
+			metav1.DeleteOptions{PropagationPolicy: &deletePolicy}); err != nil {
+			log.Info("Failed to delete build", "name", build.Name, "namespace", build.Namespace, "error", err)
```
steved, line 112: Ditto, feels like we should best effort delete as many as we can, but bubble up any errors to the exit code

pkg/cmd/controller/root.go
JadeHayes: yea I was thinking that was kinda redundant

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,78 @@
...
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          annotations:
+            checksum/config: {{ include (print $.Template.BasePath "/controller/secret.yaml") . | sha256sum }}
+            {{- if .Values.controller.metrics.enabled }}
+            prometheus.io/port: "{{ .Values.controller.manager.metricsPort }}"
+            prometheus.io/scrape: "true"
+            {{- end }}
```
steved, line 16: We don't have metrics for the cronjob so this isn't necessary

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,78 @@
...
+            {{- with .Values.controller.podAnnotations }}
+            {{- toYaml . | trimSuffix "\n" | nindent 8 }}
+            {{- end }}
+            {{- with .Values.podAnnotations }}
+            {{- toYaml . | nindent 8 }}
+            {{- end }}
+          labels:
+            {{- include "hephaestus.controller.labels.matchLabels" . | nindent 12 }}
+              {{- with .Values.controller.podLabels }}
```
steved, line 25: ultimately it doesn't matter to helm because you're using `{{- ... | nindent }}`, but I'd like the indent level to match between `matchLabels` and `podLabels`

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,78 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "hephaestus.manager.name" . }}-cronjob
```
steved, line 4: The metadata should also have labels (`labels.standard` helper)

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,78 @@
...
+              imagePullPolicy: {{ .Values.controller.manager.image.pullPolicy }}
+              volumeMounts:
+                - name: config-vol
+                  readOnly: true
+                  mountPath: /etc/hephaestus/config.yaml
+                  subPath: config.yaml
+              command: [ "hephaestus-controller" ]
+              args:
+                - run-gc
+                - --maxIBRetention=2
```
steved, line 53: hardcoded?

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,78 @@
...
+                - --config=/etc/hephaestus/config.yaml
+              {{- with .Values.controller.manager }}
+              {{- if .extraEnvVars  }}
+              env:
+                {{- with .extraEnvVars }}
+                {{- include "common.tplvalues.render" (dict "value" . "context" $) | nindent 12 }}
+                {{- end }}
+              {{ end }}
+              {{ end }}
```
steved, line 62: I think this can be simplified
```suggestion
              {{- with .Values.controller.manager.extraEnvVars  }}
              env:
                {{- include "common.tplvalues.render" (dict "value" . "context" $) | nindent 12 }}
              {{- end }}
```

deployments/helm/hephaestus/values.yaml
```diff
@@ -117,6 +117,9 @@ controller:
   # Additional volume mounts added to both the manager and vector containers
   extraVolumeMounts: []
 
+  ibGarbageCollector:
+    interval: "*/30 * * * *"
```
steved, line 121: TBH I think we can increase this further. Maybe even once or twice a day?

pkg/cmd/controller/root.go
```diff
@@ -51,6 +52,36 @@ func newStartCommand() *cobra.Command {
...
+			maxIBRetention, err := cmd.Flags().GetInt("maxIBRetention")
+			if err != nil {
+				return err
+			}
+
+			return controller.RunGC(maxIBRetention, cfg.Manager)
+		},
+	}
+	cmd.Flags().Int("maxIBRetention", 5, "Maximum number of image builds to retain."+
+		"We will retain the newest builds.")
```
steved, line 81: I like the description from forge as it's very specific:
```
Delete all ContainerImageBuild resources in a 'finished' state that exceed this count
```

pkg/controller/runGC.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+type ImageBuildGC struct {
+	maxIBRetention int
+	hephClient     clientset.Interface
+	namespaces     []string
+}
+
+func NewImageBuildGC(maxIBRetention int, log logr.Logger, ibNamespaces []string) (
+	*ImageBuildGC, error) {
```
steved, line 29: I'm very surprised this is what the go formatter decided on...

pkg/controller/runGC.go
```diff
@@ -0,0 +1,154 @@
...
+	log = log.WithName("GC")
+	ctx := context.Background()
+
+	gc, err := NewImageBuildGC(maxIBRetention, log, cfg.WatchNamespaces)
+	if err != nil {
+		log.Info("Error setting up GC", "error", err)
+		return err
+	}
+
+	log.Info("Launching Image Build GC", "maxIBRetention", gc.maxIBRetention, "namespaces", gc.namespaces)
```
steved, line 145: Let's be consistent and call it either `GC` or `Cleanup` everywhere in this file

pkg/controller/runGC.go
```diff
@@ -0,0 +1,154 @@
...
+	deletePolicy := metav1.DeletePropagationForeground
+	for _, build := range builds[:len(builds)-gc.maxIBRetention] {
+		if err := gc.hephClient.HephaestusV1().ImageBuilds(namespace).Delete(ctx, build.Name,
+			metav1.DeleteOptions{PropagationPolicy: &deletePolicy}); err != nil {
+			log.Info("Failed to delete build", "name", build.Name, "namespace", build.Namespace, "error", err)
+			errList = append(errList, err)
+		}
+		log.Info("Deleted build", "name", build.Name, "namespace", build.Namespace)
+	}
+	if len(errList) > 0 {
```
steved, line 118: Maybe https://pkg.go.dev/errors#Join?

pkg/controller/runGC_test.go
```diff
@@ -0,0 +1,197 @@
...
+				},
+			}
+			return true, obj, nil
+		})
+	FakeGC := &ImageBuildGC{
+		maxIBRetention: 1,
+		hephClient:     fakeClientErr,
+		namespaces:     []string{"aloha"},
+	}
+	err := FakeGC.CleanUpIBs(context.Background(), logr.Discard(), "aloha")
```
steved, line 165: I don't think we're testing that there is still an IB after the GC call?

pkg/controller/runGC_test.go
```diff
@@ -0,0 +1,197 @@
+package controller
```
steved, line 1: It might be nice to have a test case for the timestamp ordering: are we properly deleting the oldest X image builds?

steved (commented):
Couple minor comments, but LGTM!

deployments/helm/hephaestus/templates/controller/cronjob.yaml
JadeHayes: whoops

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,77 @@
...
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          labels:
+              {{- include "hephaestus.controller.labels.matchLabels" . | nindent 12 }}
+              {{- with .Values.controller.podLabels }}
+              {{- toYaml . | trimSuffix "\n" | nindent 12 }}
+              {{- end }}
+              {{- with .Values.podLabels }}
+              {{- toYaml . | nindent 12 }}
+              {{- end }}
```
steved, line 27: super nitpick, but can you line this up with the annotations indentation?

deployments/helm/hephaestus/templates/controller/cronjob.yaml
```diff
@@ -0,0 +1,77 @@
...
+        spec:
+          securityContext:
+            {{- toYaml .Values.controller.podSecurityContext | nindent 12 }}
+          restartPolicy: OnFailure
```
steved, line 31: Generally I'd rather see jobs have a `restartPolicy: Never` and and an explicit, configurable backoffLimit
https://kubernetes.io/docs/concepts/workloads/controllers/job/#handling-pod-and-container-failures

Otherwise, I think the job will stay failing and restarting indefinitely

deployments/helm/hephaestus/values.yaml
```diff
@@ -117,6 +117,10 @@ controller:
   # Additional volume mounts added to both the manager and vector containers
   extraVolumeMounts: []
 
+  ibGarbageCollector:
```
steved, line 120: Nitpicking again, `ibGarbageCollector` isn't the clearest name - I'd rather see something verbose like `completedImageBuildCleanup`.

deployments/helm/hephaestus/values.yaml
```diff
@@ -117,6 +117,10 @@ controller:
...
+    interval: "@daily"
+    maxIBRetention: 5
```
steved, line 122: Ditto, `maxIBRetention` isn't clear on its purpose - Forge had it as a `gc-max-keep` argument, maybe `keepMaximum: 5` or `retainMax: ...`?

pkg/controller/runGC.go
```diff
@@ -0,0 +1,154 @@
...
+	} else {
+		// create k8s client to access all namespaces in the cluster
+		k8sClient, err := k8s.NewForConfig(config)
+		if err != nil {
+			return nil, err
+		}
+		ns, err = getAllNamespaces(k8sClient)
+		if err != nil {
+			log.Info("Unable to access cluster namespaces")
+			return nil, err
```
steved, line 52: Consider using error wrapping instead of the log + return pattern:
```golang
return nil, fmt.Errorf("Unable to access cluster namespaces: %w", err)
```

pkg/controller/runGC.go
```diff
@@ -0,0 +1,154 @@
...
+		ns = append(ns, namespace.Name)
+	}
+	return ns, nil
+}
+
+func (gc *ImageBuildGC) GCImageBuilds(ctx context.Context, log logr.Logger, namespace string) error {
+	crdList, err := gc.hephClient.HephaestusV1().ImageBuilds(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		log.Info("Unable to access a list of IBs, not starting Image Build GC", "error", err)
+		return err
```
steved, line 80: Same error wrapping comment

pkg/controller/runGC.go
```diff
@@ -0,0 +1,154 @@
...
+		return builds[i].CreationTimestamp.Before(&builds[j].CreationTimestamp)
+	})
+
+	var errList []error
+	deletePolicy := metav1.DeletePropagationForeground
+	for _, build := range builds[:len(builds)-gc.maxIBRetention] {
+		if err := gc.hephClient.HephaestusV1().ImageBuilds(namespace).Delete(ctx, build.Name,
+			metav1.DeleteOptions{PropagationPolicy: &deletePolicy}); err != nil {
+			log.Info("Failed to delete build", "name", build.Name, "namespace", build.Namespace, "error", err)
+			errList = append(errList, err)
```
steved, line 113: Wrapping this error and adding the build name will help debugging later

pkg/controller/runGC.go
```diff
@@ -0,0 +1,154 @@
...
+	sort.Slice(builds, func(i, j int) bool {
+		return builds[i].CreationTimestamp.Before(&builds[j].CreationTimestamp)
+	})
+
+	var errList []error
+	deletePolicy := metav1.DeletePropagationForeground
+	for _, build := range builds[:len(builds)-gc.maxIBRetention] {
+		if err := gc.hephClient.HephaestusV1().ImageBuilds(namespace).Delete(ctx, build.Name,
+			metav1.DeleteOptions{PropagationPolicy: &deletePolicy}); err != nil {
+			log.Info("Failed to delete build", "name", build.Name, "namespace", build.Namespace, "error", err)
```
steved, line 112: Is this log line redundant with the returned error list that's logged when the program exits?

pkg/controller/runGC.go
```diff
@@ -0,0 +1,154 @@
...
+	log := ctrlzap.New(
+		ctrlzap.UseDevMode(true),
+		ctrlzap.Encoder(zapcore.NewConsoleEncoder(zap.NewDevelopmentEncoderConfig())),
+	)
+	log = log.WithName("GC")
+	ctx := context.Background()
+
+	gc, err := NewImageBuildGC(maxIBRetention, log, cfg.WatchNamespaces)
+	if err != nil {
+		log.Info("Error setting up GC", "error", err)
```
steved, line 141: Same error wrapping comment

pkg/controller/runGC.go
```diff
@@ -0,0 +1,154 @@
...
+		return err
+	}
+
+	log.Info("Launching Image Build GC", "maxIBRetention", gc.maxIBRetention, "namespaces", gc.namespaces)
+	for _, ns := range gc.namespaces {
+		err := gc.GCImageBuilds(ctx, log, ns)
+		if err != nil {
+			log.Info("Exiting Image Builder GC due to error: ", "error", err)
+			os.Exit(1)
```
steved, line 150: Why exit instead of return?

pkg/controller/runGC_test.go
```diff
@@ -0,0 +1,256 @@
...
+
```
steved, line 2: I don't think I see a test where we're making sure the GC doesn't touch non-failed/succeeded builds (i.e. those running or pending)

steved (approved):
Couple nitpicks and minor comments, but it LGTM

PR #90, opened 2023-03-28
Helpers for using rootless with non-default UIDs

Secretions:
Using other UIDs should not really be recommended, due to the custom image, but in the event that it's necessary (which one organization is speculating it may be for them), this will give some template to follow.

### DIFF
--- a/build/buildkit/Dockerfile
+++ b/build/buildkit/Dockerfile
@@ -13,7 +13,8 @@ RUN apk update && \
     apk upgrade && \
     rm -rf /var/cache/apk
 
-RUN deluser user || echo this fails for no reason && \
+# You can ignore "deluser: can't find user in /etc/group"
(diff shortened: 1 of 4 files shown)


build/buildkit/Dockerfile
```diff
@@ -7,12 +7,23 @@ RUN apk update && \
 
 FROM moby/buildkit:${BUILDKIT_TAG} AS rootless
 ARG ISTIO_GID=1337
+ARG UID=1000
 USER root
 RUN apk update && \
     apk upgrade && \
     rm -rf /var/cache/apk
+
+RUN deluser user || echo this fails for no reason && \
```
steved, line 16: As far as I can tell the `deluser` message is just informational and it's not failing

steved (approved):

PR #93, opened 2023-04-10
cleanup cronjob must support istio during startup

steved:
To properly work with Istio, the cronjob must be able to wait for the proxy sidecar to start and then call `/quitquitquit` at exit.


JadeHayes (approved):
LGTM thank you!  Didn't realize istio needed special configuration

PR #95, opened 2023-04-24
Updates

fraenkel:


Dockerfile
```diff
@@ -9,7 +9,7 @@ COPY deployments/crds ./deployments/crds
 ENV CGO_ENABLED=0 GOOS=linux
 RUN go build -o hephaestus-controller ./cmd/controller
 
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static:debug-nonroot
```
ddl-dclegg, line 12: Why `debug` and not just `non-root`?

ddl-dclegg (commented):
Things look good generally, I just have the one question on the static image choice.

Dockerfile
fraenkel: that will be reverted... when debugging... I needed access to a shell to see what is going on.

fraenkel: reverted

ddl-dclegg (approved):

PR #97, opened 2023-05-18
DOM-46835 - Vector_ metrics have significantly high cardinality and are showing to be very expensive

namudomino:
1. Exploded after enabling prometheus remote_write with default interval to 5seconds.
2. 120$ per month from data ingest on vector_ to newrelic.


deployments/helm/hephaestus/templates/controller/secret.yaml
```diff
@@ -107,6 +107,7 @@ stringData:
         inputs:
           - internal_metrics
         endpoint: https://metric-api.newrelic.com/prometheus/v1/write?prometheus_server={{ .Values.newRelic.appName }}
+        interval: {{ .Values.newRelic.remoteWriteInterval }}
```
steved, line 110: Does this work? I don't see `interval` anywhere, only `batch.timeout_secs`

namudomino: This can be  on hold for now until get an answer from to newrelic how to configure this same as the default scrape window we have in our prometheus crape which is 2m.  (default remote_write is 5s which is going to do frrequent data ingestion to NR) 

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write doesnt have any option for setting this  . Looks like delaying prometheus remote_write may lead to  missing data writes or performance issues as its streamed from memory buffer to newrelic.

Borrowed the above setting  used in the pr from an example https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.QueueConfig or some place when exploring options on how to set it.

namudomino: I m closing this pr for now as drop rules took care of cost reduction and adding configuration on prometheus_rewrite to add interval is not feasible by design.

PR #100, opened 2023-06-12
PLAT-5981: clarify error message when registry validation fails on an image build

ddl-kfrench:
as we can't establish a definitive 1:1 correlation between the server field in registry auth and the computed docker config.json in downstream authentication, this adds a `helpMessage` str slice to store general metadata about the credentials supplied to an `imagebuild` that can be leveraged in any error message(s) related to failed registry authentication.

### DIFF
--- a/pkg/controller/support/credentials/credentials.go
+++ b/pkg/controller/support/credentials/credentials.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/registry"
@@ -43,13 +44,19 @@ func Persist(
 	logger logr.Logger,
 	cfg *rest.Config,
 	credentials []hephv1.RegistryCredentials,
-) (string, error) {
+) (string, []string, error) {
 	dir, err := os.MkdirTemp("", "docker-config-")
 	if err != nil {
-		return "", err
+		return "", nil, err
 	}
 
 	auths := AuthConfigs{}
+	// as we can't establish a 1:1 correlation between the server field
+	// and the computed docker config.json in downstream authentication
+	// helpMessage stores general meta-information about the creds
+	// in use that can be supplied to any error message(s) that surface
+	// for more easily debugging the source of a failed auth.
+	var helpMessage []string
 	for _, cred := range credentials {
 		var ac types.AuthConfig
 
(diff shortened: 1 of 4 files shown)


pkg/controller/support/credentials/credentials.go
```diff
@@ -136,7 +147,9 @@ func Verify(ctx context.Context, configDir string, insecureRegistries []string)
 		auth.ServerAddress = server
 
 		if _, _, err = svc.Auth(ctx, &auth, "DominoDataLab_Hephaestus/1.0"); err != nil {
-			errs = append(errs, fmt.Errorf("%q client credentials are invalid: %w", server, err))
+			//nolint:lll
+			detailedErr := fmt.Errorf("%q client credentials are invalid. Make sure the provided credentials are correct: %s. Underlying error: %w", server, strings.Join(helpMessage, ","), err)
```
steved, line 151: Not 100% sure how the formatting looks here, so apologies if it's weird. Just looking for a bit more readability for users.
```suggestion
			detailedErr := fmt.Errorf("unable to authenticate to registry %q\nMake sure the following sources of credentials are correct: %s\nUnderlying error: %w", server, strings.Join(helpMessage, ", "), err)
```

or maybe

```golang
detailedErr := fmt.Errorf("unable to authenticate to registry %q\nMake sure the following sources of credentials are correct: %s", server, strings.Join(helpMessage, ", "))
errs = append(errs, detailedErr, err)
```

pkg/controller/support/credentials/credentials.go
```diff
@@ -43,57 +44,67 @@ func Persist(
...
 			if err := json.Unmarshal(secret.Data[corev1.DockerConfigJsonKey], &conf); err != nil {
-				return "", err
+				return "", nil, err
 			}
 
 			for server, config := range conf.Auths {
 				auths[server] = config
 			}
 
+			helpMessage = append(helpMessage, fmt.Sprintf("secret %q in namespace %q", cred.Secret.Name, cred.Secret.Namespace))
```
steved, line 89: I wonder if this should contain which servers it has credentials for or is that now info overload?

```golang
var servers []string
for server, config := range conf.Auths {
  auths[server] = config
  servers = append(servers, server) 
}

helpMessage = append(helpMessage, fmt.Sprintf("secret %q in namespace %q (credentials for servers: %s)", cred.Secret.Name, cred.Secret.Namespace, strings.Join(servers, ", ")))
```

pkg/controller/support/credentials/credentials.go
```diff
@@ -43,57 +44,67 @@ func Persist(
...
+			helpMessage = append(helpMessage, "basic authentication username and password")
 		case pointer.BoolDeref(cred.CloudProvided, false):
 			pac, err := CloudAuthRegistry.RetrieveAuthorization(ctx, logger, cred.Server)
 			if err != nil {
-				return "", fmt.Errorf("cloud registry authorization failed: %w", err)
+				return "", nil, fmt.Errorf("cloud registry authorization failed: %w", err)
 			}
 
 			ac = *pac
+			helpMessage = append(helpMessage, "cloud provider install configuration in image_building.cloud_registry_auth")
```
steved, line 105: On further thought, this is probably a bit too domino-specific.
```suggestion
			helpMessage = append(helpMessage, "cloud provider access configuration")
```

steved (approved):
Mostly nitpicks around wording so merge when you feel comfortable. Thanks!

PR #104, opened 2023-06-16
(fix): creds unit test

ddl-kfrench:
resolves an issue in the unit test for validating the credentials error


fraenkel (approved):
you could test that the servers are in the message somewhere too.

steved (approved):

PR #105, opened 2023-06-26
[PLAT-6701] Enable zstd hephaestus

ddl-retornam:
This pr creates the options to enable `zstd` and `estargz` compression.


pkg/buildkit/buildkit.go
```diff
@@ -118,6 +119,25 @@ type Client struct {
 	dockerAuthConfig string
 }
 
+func validateCompression(compression hephconfig.BuildkitCompression, name string) map[string]string {
```
ddl-kfrench, line 122: we can leverage the compression string types we define in the cfg package here ->

```
func validateCompression(compression hephconfig.BuildkitCompression, name string) map[string]string {
	attrs := map[string]string{
		"name": name,
	}

	switch compression {
	case hephconfig.CompressionStargz:
		attrs["push"] = "true"
		attrs["compression"] = "estargz"
		attrs["force-compression"] = "true"
		attrs["oci-mediatypes"] = "true"
	case hephconfig.CompressionZstd:
		attrs["compression"] = "zstd"
		attrs["force-compression"] = "true"
	case hephconfig.CompressionGzip:
		attrs["push"] = "true"
	}

	return attrs
}
```

also, more of a nit/naming thing but it might be better to use something like `setCompressionAttrs` since we're not performing validation in any strict sense 

we could extend that example above more if we wanted some additional validation and have `default` return an error for unsupported compression types

pkg/config/config.go
```diff
@@ -52,6 +55,13 @@ func (c Controller) Validate() error {
 		errs = append(errs, "newRelic.licenseKey cannot be blank")
 	}
```
ddl-kfrench, line 57: i don't think we need to use `lo` here - we can just use our faux-go-enum types as a validation

```
	switch c.Buildkit.Compression {
	case CompressionGzip, CompressionStargz, CompressionZstd:
		break
	default:
		// force gzip as the default if not otherwise set
		c.Buildkit.Compression = CompressionGzip
		CompressionMethod = CompressionGzip
	}
```

pkg/buildkit/buildkit.go
retornam: fixed in latest update

pkg/config/config.go
ddl-retornam: resolved

pkg/config/config.go
```diff
@@ -52,6 +51,13 @@ func (c Controller) Validate() error {
...
 
+	switch c.Buildkit.Compression {
+	case CompressionGzip, CompressionStargz, CompressionZstd:
+		CompressionMethod = c.Buildkit.Compression
+	default:
+		CompressionMethod = CompressionGzip
```
ddl-kfrench, line 58: do we need to ensure that gzip is forced as the default if not otherwise set? i.e.

```
	switch c.Buildkit.Compression {
	case CompressionGzip, CompressionStargz, CompressionZstd:
		CompressionMethod = c.Buildkit.Compression
	default:
		// force gzip as the default if not otherwise set
		c.Buildkit.Compression = CompressionGzip
		CompressionMethod = CompressionGzip
	}
```

ddl-kfrench (approved):
lgtm - just had one small note

PR #117, opened 2023-10-18
PLAT-7461: Update dependencies

fraenkel:
regen code
fix lint issues


deployments/crds/hephaestus.dominodatalab.com_imagebuilds.yaml
```diff
@@ -3,8 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.3
-  creationTimestamp: null
```
Secretions, line 7: Do we know why this was here?

pkg/controller/start.go
```diff
@@ -102,29 +103,33 @@ func createManager(log logr.Logger, cfg config.Manager) (ctrl.Manager, error) {
...
 
 	if certDir := os.Getenv("WEBHOOK_SERVER_CERT_DIR"); certDir != "" {
 		log.Info("Overriding webhook server certificate directory", "value", certDir)
-		opts.WebhookServer = &webhook.Server{
-			CertDir: certDir,
-		}
+		webhookOpts.CertDir = certDir
 	}
 
+	restConfig := ctrl.GetConfigOrDie()
```
Secretions, line 118: Why are we setting this here now?

fraenkel: sorry, it can move back. The old code did something weird and used the cacheFunc instead of the Cache options.
I will move it back since we won't be using cache.New().

deployments/crds/hephaestus.dominodatalab.com_imagebuilds.yaml
fraenkel: its all code generation.
guess it was a think in the older code gen but not anymore.

PR #119, opened 2023-10-23
GC runs as part of the controller

fraenkel:
Run image build GCs every hour


pkg/controller/imagebuild/imagebuild.go
```diff
@@ -15,22 +15,35 @@ import (
 	"github.com/dominodatalab/hephaestus/pkg/controller/imagebuild/predicate"
 )
 
-var ch = make(chan client.ObjectKey)
-
-func Register(mgr ctrl.Manager, cfg config.Controller, pool worker.Pool, nr *newrelic.Application) error {
-	return core.NewReconciler(mgr).
+func Register(mgr ctrl.Manager, cfg config.Controller, pool worker.Pool, nr *newrelic.Application,
+	deleteChan chan client.ObjectKey,
+) error {
```
steved, line 20: Very nitpicky, but this formatting situation feels wrong

pkg/controller/imagebuild/component/imagebuildgc.go
```diff
@@ -0,0 +1,112 @@
...
+		if state == hephv1.PhaseFailed || state == hephv1.PhaseSucceeded {
+			builds = append(builds, ib)
+		}
+	}
+
+	if len(builds) <= gc.HistoryLimit {
+		return nil
+	}
+
+	logger.Info("Removing ImageBuilds", "count", len(builds))
```
steved, line 83: This should be on line 98 for an accurate count right?

pkg/controller/imagebuild/component/imagebuildgc.go
```diff
@@ -0,0 +1,112 @@
...
+	sort.Slice(builds, func(i, j int) bool {
+		iTS := builds[i].CreationTimestamp
+		jTS := builds[j].CreationTimestamp
+		if before := iTS.Before(&jTS); before {
```
steved, line 87: more curious than a style thing, but why not?
```suggestion
		if iTS.Before(&jTS) {
```

pkg/controller/imagebuild/component/imagebuildgc.go
```diff
@@ -0,0 +1,112 @@
...
+			return true
+		}
+		if iTS.Equal(&jTS) {
+			if builds[i].Namespace < builds[j].Namespace {
+				return true
+			}
+			return builds[i].Name < builds[j].Name
+		}
+		return false
+	})
```
steved, line 97: I'm not sure it's really an issue, but this would keep `5` image builds globally vs. `5` per namespace before.

steved (approved):
Minor things but looks good

deployments/helm/hephaestus/values.yaml
```diff
@@ -126,8 +121,10 @@ controller:
       pullPolicy: IfNotPresent
       pullSecrets: []
 
-    # Maximum number of concurrent builds which can be run
-    imageBuildMaxConcurrency: 5
```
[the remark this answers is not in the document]
fraenkel: why, you would have to update the image without the chart. which we would really want to catch.
But I can add it...

pkg/controller/imagebuild/imagebuild.go
fraenkel: There isn't a good way...  I can go with a long line but then the linter complains... let me see if I can force this to work well with gofumpt. But the pipelines does a gofmt so its a sad state of affairs.

pkg/controller/imagebuild/component/imagebuildgc.go
fraenkel: its what was there... I will move it down and change the key to "removing" or something since count is not clear.

pkg/controller/imagebuild/component/imagebuildgc.go
fraenkel: because I was doing something else until I wasn't

pkg/controller/imagebuild/component/imagebuildgc.go
fraenkel: so you would have thought that, but its not how it works today. And I only stumbled on it by testing for reals. but yes it makes less sense.
In our case it doesn't matter given we only have 1 namespace.

pkg/controller/imagebuild/component/imagebuildgc.go
steved: The message was a bit different ("Total resources eligible for deletion")

deployments/helm/hephaestus/values.yaml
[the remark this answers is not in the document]
steved: I guess I was thinking of someone who just did (wrongly, since it's a top-level)
```
release_overrides:
  hephaestus:
    chart_values:
      controller:
        imageBuildMaxConcurrency: 10
```
Maybe the better alternative is a `fail`?

pkg/controller/imagebuild/component/imagebuildgc.go
steved: Hm, I read the current behavior as looping over `GCImageBuilds` for each NS. Was there a bug there?

deployments/helm/hephaestus/values.yaml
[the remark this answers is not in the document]
fraenkel: I think fail is good.

pkg/controller/imagebuild/component/imagebuildgc.go
fraenkel: So we default to ns = "" which is all namespaces, so we didn't split it apart and then work per namespace.

deployments/helm/hephaestus/values.yaml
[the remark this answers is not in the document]
fraenkel: extra fields will cause a failure.

pkg/controller/imagebuild/component/imagebuildgc.go
fraenkel: its at line 99,
	logger.Info("Deleting ImageBuilds", "imageBuildsToRemove", len(builds))

pkg/controller/imagebuild/imagebuild.go
fraenkel: made it consistent with arg per line.

PR #124, opened 2023-11-15
PLAT-7603 - Updated chart

ddl-dclegg:
- Bump chart version to `0.7.2`


fraenkel (changes requested):
Not how this works.
It gets stamped when the tag is created.

ddl-dclegg:
> [Nov 16, 2023 ](#pullrequestreview-1734815086)
> [ View reviewed changes ](/dominodatalab/hephaestus/pull/124/files/6670ca68af80223db8dd74f56667b62a7a2bb3c3)

@fraenkel The problem is that it _didn't_ happen in #123. If it's preferred that I close this and someone look into _why_ it wasn't versioned at merge-to-main, I will do that.

PR #126, opened 2023-11-20
Fix worker pool

fraenkel:


pkg/buildkit/worker/pool_test.go
```diff
@@ -390,39 +325,26 @@ func TestPoolGetFailedScaleUp(t *testing.T) {
...
-
-	leaseChannel := make(chan result)
-	go func() {
-		addr, err := wp.Get(ctx, owner)
-		leaseChannel <- result{addr, err}
-	}()
+	expected := "tcp://buildkit-0.buildkit.test-namespace:1234"
+	if addr != expected {
+		t.Errorf("did not received correct lease: %s expected, %s actual", expected, addr)
+	}
```
steved, line 340: IMO this isn't the same test, though I can understand why this is changed. I think the intention was:

1. Lease
2. Lease
3. Release number 1
4. Validate lease number 2 received the existing address

pkg/buildkit/worker/pool_test.go
```diff
@@ -1028,62 +922,62 @@ func TestPoolPodReconciliation(t *testing.T) {
...
+//
+// 	t.Run("Close twice", func(t *testing.T) {
+// 		wp := NewPool(fakeClient, testConfig, SyncWaitTime(10*time.Millisecond), MaxIdleTime(5*time.Minute), Logger(testr.New(t)))
+// 		if err := wp.Start(context.Background()); err != nil {
+// 			t.Error(err)
+// 		}
+// 		wp.Close()
+// 		wp.Close()
+// 	})
+// }
```
steved, line 980: Guess these are removable?

pkg/controller/start.go
```diff
@@ -111,14 +111,15 @@ func createManager(log logr.Logger, cfg config.Manager) (ctrl.Manager, error) {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(hephv1.AddToScheme(scheme))
 
+	shutdownTimeout := 5 * time.Minute
```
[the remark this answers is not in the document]
fraenkel: change?
I am planning to move this into config but I just want to test out my theory.

pkg/buildkit/worker/pool_test.go
fraenkel: Was focused on fixing Get, I forgot I had more tests to fix/delete.

pkg/buildkit/worker/pool_test.go
fraenkel: See it now. Will fix it.

fraenkel: Although the test is testing whether it failed the scale up, not get/release/get which is different.

fraenkel: This test makes less and less sense.
There is no validation that a scale up was even attempted given the sts is set to 0 replicas.
A get/release/get test should be separate and I will add it to TestRelease since its part of testing the expected behavior.

fraenkel: Ok, I see what is supposed to happen... Will add more verification of expectations which will make this more clear.

pkg/controller/start.go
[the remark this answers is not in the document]
steved: sorry, wording was a bit unclear. you added it in deployment.yaml.

steved (approved):
Just the commented out tests in pool_test.go.

PR #131, opened 2024-01-04
[DOM-49106] Emit image size as part of build

ddl-rliu:
## Problem

When an image is built, we do not track the size of the image

## Solution

Track the size and return it as `compressedImageSizeBytes`, part of the `Annotations` field in the `BuildSucceeded` response message at the end of a build.

We do this in Hephaestus, rather than Nucleus. This is because Nucleus does not talk with the registry; rather, it sets up the registry configs/auth and defers the responsibility to Hephaestus (which then defers it to buildkit). This PR queries the registry in order to track the size, and Hephaestus seems preferable since it is "closer" to buildkit.

https://dominodatalab.atlassian.net/browse/DOM-49106

1. This PR does not handle multi-arch images. This (and the other below item) is tracked in a followup ticket https://dominodatalab.atlassian.net/browse/DOM-53850
2. This PR only handles Hephaestus images.


pkg/buildkit/buildkit.go
```diff
@@ -354,19 +377,50 @@ func (c *Client) runSolve(ctx context.Context, so bkclient.SolveOpt) error {
...
+		img, err := remote.Image(ref, remote.WithContext(ctx), remote.WithAuth(auth))
+		if err != nil {
+			return err
+		}
+		layers, err := img.Layers()
+		if err != nil {
+			return err
+		}
+		for _, layer := range layers {
+			compressedSize, err := layer.Size()
```
ddl-rliu, line 409: This is the size of the compressed layer. The uncompressed layer size could probably be calculated by reading in the response from `layer.Uncompressed()`.

pkg/controller/imagebuild/component/builddispatcher.go
```diff
@@ -253,6 +256,13 @@ func (c *BuildDispatcherComponent) Reconcile(coreCtx *core.Context) (ctrl.Result
 	obj.Status.BuildTime = time.Since(start).Truncate(time.Millisecond).String()
 	buildSeg.End()
 
+	log.Info("Final image size: ", "imageSize", imageSize)
+	annotations := obj.GetAnnotations()
+	annotations["imagebuilder.dominodatalab.com/image-size"] = strconv.FormatInt(imageSize, 10)
+	obj.SetAnnotations(annotations)
```
ddl-rliu, line 262: Trying to figure out why this annotation doesn't appear in https://github.com/dominodatalab/hephaestus/pull/131/files#diff-5c6176a5f8615022b3d6cd35ca34aaa3bc316e6c27af6c35037080ee2ff4142eR205

ddl-rliu: annotations on the object in builddispatcher
```
{
    "imagebuilder.dominodatalab.com/build-id": "6597360e87e5c341bde190fd",
    "imagebuilder.dominodatalab.com/image-size": "3549772587",
    "imagebuilder.dominodatalab.com/last-image": "153827342122.dkr.ecr.us-west-2.amazonaws.com/richard31773/environment:65971a0e87e5c341bde190bb-9",
    "imagebuilder.dominodatalab.com/version": "v3"
}
```
in amqpmessenger
```
{
    "imagebuilder.dominodatalab.com/build-id": "6597360e87e5c341bde190fd",
    "imagebuilder.dominodatalab.com/last-image": "153827342122.dkr.ecr.us-west-2.amazonaws.com/richard31773/environment:65971a0e87e5c341bde190bb-9",
    "imagebuilder.dominodatalab.com/version": "v3"
}
```
both objects have the same uid and name
```
uid: "a3a70f97-7a90-46d8-b674-038bb17a4670",
name: "domino-build-65981a4dd9be8407b9c74e3b"
```

pkg/controller/imagebuildmessage/component/amqpmessenger.go
```diff
@@ -201,6 +202,7 @@ func (c *AMQPMessengerComponent) Reconcile(ctx *core.Context) (ctrl.Result, erro
 				images = append(images, reference.TagNameOnly(named).String())
 			}
 			message.ImageURLs = images
+			message.Annotations["imagebuilder.dominodatalab.com/compressed-image-size"] = ib.Status.CompressedImageSize
```
ddl-ryan-connor, line 205: can we include the units in the name of the annotation and go var name? e.g. `bytes`

pkg/buildkit/buildkit.go
```diff
@@ -354,19 +377,49 @@ func (c *Client) runSolve(ctx context.Context, so bkclient.SolveOpt) error {
...
 		}
 
 		c.log.Info("Solve complete")
+		imageName := res.ExporterResponse["image.name"]
+		ref, err := name.ParseReference(imageName)
+		if err != nil {
+			return err
+		}
+		registryName := ref.Context().RegistryStr()
+		auth, err := c.ResolveAuth(registryName)
```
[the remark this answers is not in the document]
ddl-rliu: Yeah, the thinking here was that - here, `buildkit.Solve` returns a response, but that response unfortunately doesn't include the final image size - so we use the image url and calculate the image size to return (sum the layers).

For reference, the solve response looks approximately like
```
{
    "containerimage.config.digest": ...,
    "containerimage.descriptor": {
        "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
        "digest": "...",
        "size": 2267,
        "platform": {
            "architecture": "amd64",
            "os": "linux"
        }
    },
    "containerimage.digest": "...",
    "image.name": "..."
}
```

pkg/controller/imagebuildmessage/component/amqpmessenger.go
ddl-rliu: done

pkg/buildkit/buildkit.go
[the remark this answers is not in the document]
ddl-rliu: Good point, I was also worried about adding the new dependencies into Heph - I think I'll schedule a quick call after hackathon for the team to think through these different options.

Re: failure modes - that makes sense, just pushed a change that improves this a little bit https://github.com/dominodatalab/hephaestus/pull/131/files#diff-9bf6bd7b5671bcb828a58b3c5ef62d040d8ded0f7dc901e2b4c01191cb025dd4R418-R421 - the errors when calculating image size should _**not**_ count as a build error (it's metadata) - instead, this will log any errors and still return "build success".

TIL about the multi-arch nuance, I'll have to think about that.

pkg/buildkit/buildkit.go
```diff
@@ -144,11 +147,11 @@ func validateCompression(compression string, name string) map[string]string {
 	return attrs
 }
 
-func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
+func (c *Client) Build(ctx context.Context, opts BuildOptions) (int64, error) {
```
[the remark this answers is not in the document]
ddl-rliu: my thinking here was that `buildkit.go::Build` is like a wrapper around `buildkit/solve.go::Solve`, and it was odd that `Solve` returns `(*SolveResponse, error)` while `Build` returns `error`. An odd thing about my change is that the entirety of the `Build` response is an int64 of the image size, but maybe its reasonable.

I'm not all that familiar with Go interfaces, so I'll review https://go.dev/doc/effective_go#interfaces in the meantime and see what I find.

Secretions (approved):

pkg/buildkit/buildkit.go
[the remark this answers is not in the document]
ddl-rliu: Another option is to return a BuildResult struct here, perhaps like
```
type BuildResult struct {
   Size int64
   Err error
}
```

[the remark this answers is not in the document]
ddl-ryan-connor: seems like a convention in go is to return `err` as the last return val so i dont think putting it inside a struct would be better

pkg/controller/imagebuild/component/builddispatcher.go
ddl-rliu: 1) Need to update structs
2) Need to run `make crds && make apply`

PR #134, opened 2024-02-23
[DOM-53893] Parse image labels

ddl-rliu:
https://dominodatalab.atlassian.net/browse/DOM-53893

Changes to Hephaestus to Parse container image labels and add them to annotations on the ImageSuceeded message.

Hephaestus sends "ImageSucceeded" to Nucleus with some metadata.

Metadata before changes:

```
imagebuilder.dominodatalab.com/build-id -> 65de3c0c1c7378567a552ad2
imagebuilder.dominodatalab.com/compressed-image-size-bytes -> 2482008151
imagebuilder.dominodatalab.com/last-image -> 153827342122.dkr.ecr.us-west-2.amazonaws.com/richard38490/environment:65d92a2efe46a21d14bd7899-19
imagebuilder.dominodatalab.com/version -> v3
...
```

Metadata after changes (net new, this includes the labels on the container image):

```
(same as above PLUS)
maintainer -> Domino Data Lab <support@dominodatalab.com>
org.opencontainers.image.description -> Standard Domino environment for data science
org.opencontainers.image.ref.name -> ubuntu
org.opencontainers.image.title -> Domino Standard Environment
org.opencontainers.image.vendor -> Domino Data Lab
com.dominodatalab.environment.workspace-properties -> anVweXRlc...YWxzZQ==			
...
```


deployments/crds/hephaestus.dominodatalab.com_imagebuilds.yaml
```diff
@@ -222,6 +222,12 @@ spec:
                   - type
                   type: object
                 type: array
+              labels:
+                additionalProperties:
+                  type: string
+                description: 'Map of string keys and values roughly corresponding
+                  to Kubernetes Labels More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels'
```
noahjax, line 229: Not sure this is quite accurate, the image labels here for the most part don't map to k8s labels. It's still valuable to fetch and store them though, as some do and others (i.e. workspace properties) may be useful for other reasons.

ddl-rliu: Right, then I've got it mixed up with OCI image config labels. Good catch, updated the description.

ddl-rliu: https://github.com/dominodatalab/hephaestus/pull/134/files#diff-46b29db23c74cff80af70cc51499e89f78828c132fa4bbf743a0a684fe003522R228-R229

> Map of string keys and values corresponding OCI image config labels. Labels contains arbitrary metadata for the container

pkg/controller/imagebuild/component/builddispatcher.go
```diff
@@ -231,7 +235,7 @@ func (c *BuildDispatcherComponent) Reconcile(coreCtx *core.Context) (ctrl.Result
 
 	// best effort phase change regardless if the original context is "done"
 	coreCtx.Context = context.Background()
-	imageSize, err := bk.Build(buildCtx, buildOpts)
+	imageName, err := bk.Build(buildCtx, buildOpts)
```
noahjax, line 238: So you can get image size from the build itself but need to fetch the image post-build in order to get additional information like labels?

noahjax (approved):
Excited to have this, one step closer to my dream of environments auto-populating their supported tools

noahjax:
@niole (cc @ddl-ikoulinitch) Richard's work here would unblock https://dominodatalab.atlassian.net/browse/DOM-36237, which would make users lives a lot easier

pkg/buildkit/buildkit.go
```diff
@@ -414,20 +384,16 @@ func (c *Client) runSolve(ctx context.Context, so bkclient.SolveOpt) (int64, err
...
 
 	if err := eg.Wait(); err != nil {
 		c.log.Info(fmt.Sprintf("Build failed: %s", err.Error()))
-		return 0, fmt.Errorf("buildkit solve issue: %w", err)
+		return "", fmt.Errorf("buildkit solve issue: %w", err)
 	}
 
-	c.log.Info(fmt.Sprintf("Final image size: %d", size))
-	return size, nil
+	c.log.Info(fmt.Sprintf("Final image name: %s", imageName))
```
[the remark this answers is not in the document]
ddl-rliu: Here is where in `builddispatcher.go` we now log image size https://github.com/dominodatalab/hephaestus/pull/134/files#diff-858da12a46ea7b374d3935c612faa944aec57f6d6e118bf800e74b1927ce3313R282

Yep, I did check 👍

(For reference, I introduced the int64 size return value in https://github.com/dominodatalab/hephaestus/pull/131/files#diff-9bf6bd7b5671bcb828a58b3c5ef62d040d8ded0f7dc901e2b4c01191cb025dd4 )

PR #140, opened 2024-03-22
add dockerfileContents to `ImageBuild` CRD

dmcwhorter-ddl:


pkg/api/hephaestus/v1/imagebuild_webhook.go
```diff
@@ -42,12 +42,16 @@ func (in *ImageBuild) validateImageBuild(action string) (admission.Warnings, err
...
 
-	if strings.TrimSpace(in.Spec.Context) == "" {
-		log.V(1).Info("Context is blank")
-		errList = append(errList, field.Required(fp.Child("context"), "must not be blank"))
-	} else if _, err := url.ParseRequestURI(in.Spec.Context); err != nil {
-		log.V(1).Info("Context is not a valid URL")
-		errList = append(errList, field.Invalid(fp.Child("context"), in.Spec.Context, err.Error()))
+	if strings.TrimSpace(in.Spec.Context) == "" && strings.TrimSpace(in.Spec.DockerfileContents) == "" {
+		log.V(1).Info("Context and DockerfileContents are both blank")
+		errList = append(errList, field.Required(fp.Child("context or dockerFileContents"), "must not be blank"))
```
steved, line 47: this should probably be two separate errors. with `fp.Child`, it looks like:
```
The ImageBuild "dkrrm56j" is invalid: spec.context or dockerFileContents: Required value: must not be blank
```

pkg/buildkit/buildkit.go
```diff
@@ -165,18 +168,30 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) (string, error) {
...
 		}
-
 		contentsDir = extract.ContentsDir
+	case strings.TrimSpace(opts.DockerfileContents) != "":
+		c.log.Info("Creating context from DockerfileContents")
+		contentsDir, err = os.MkdirTemp(buildDir, "dockerfile-contents-")
+		if err != nil {
+			return "", fmt.Errorf("cannot create temp directory for dockerfileContents: %w", err)
+		}
+		err = os.WriteFile(path.Join(contentsDir, "Dockerfile"), []byte(opts.DockerfileContents), os.FileMode(0755))
```
steved, line 189: 0755 seems a bit much, maybe 0644?

pkg/controller/imagebuild/component/builddispatcher.go
```diff
@@ -278,9 +279,15 @@ func populateBuildStatus(obj *hephv1.ImageBuild, log logr.Logger, img v1.Image,
...
 	}
+	digest, err := img.Digest()
+	if err != nil {
+		log.Error(err, "Cannot retrieve image digest", "imageName", imageName)
+	}
 
 	log.Info(fmt.Sprintf("Final image size: %d", imageSize))
+
 	obj.Status.CompressedImageSizeBytes = strconv.FormatInt(imageSize, 10)
+	obj.Status.Digest = digest.String()
```
steved, line 290: Change is fine, just curious how this will be used?

steved (approved):

pkg/controller/imagebuild/component/builddispatcher.go
dmcwhorter-ddl: I was thinking we could use this to verify that the image produced by this ImageBuild and the one in the remote repository are actually the same image.  This would help for example when we go to delete the remote image to make sure its nothing else has pushed a newer one.

steved: Ok. Tags are immutable for Domino repos in AWS, but yes that won't be the case everywhere: 
https://github.com/dominodatalab/terraform-aws-eks/blob/882e01029ef85e276af000b2e1c21e3532d02850/modules/infra/submodules/storage/ecr.tf#L9

> This would help for example when we go to delete the remote image to make sure its nothing else has pushed a newer one.

When do we delete images?

pkg/api/hephaestus/v1/imagebuild_webhook.go
dmcwhorter-ddl: Ok I changed this around to make it more correct and say:
`The ImageBuild "dkrrm56j" is invalid: spec.context: Required value: must not be blank if spec.dockerfileContents is blank`

pkg/controller/imagebuild/component/builddispatcher.go
dmcwhorter-ddl: discussed in slack

PR #163, opened 2024-04-15
PLAT-8277: CRD jobs do not need istio

fraenkel:


ddl-dclegg (approved):
I'd just like to note that the ticket detail would've been helpful here, since dev-offline being a good validation flavor would matter later on. A description or the ticket link is nice to have. 
https://dominodatalab.atlassian.net/browse/PLAT-8277

deployments/helm/hephaestus/templates/crd-hooks.yaml
```diff
@@ -93,9 +94,6 @@ spec:
           imagePullPolicy: {{ .Values.controller.manager.image.pullPolicy }}
           args:
             - crd-apply
-            {{- if .Values.istio.enabled }}
-            - --istio-enabled
```
steved, line 97: should we remove support for this flag as a follow-up?

fraenkel: either remove it or fix it.
it doesn't handle the case where istio is enabled and there is no sidecar. Basically it shouldn't care and just deal with the case when it does find a sidecar.

fraenkel: realize this is broken in all places that copied this code, ie, 2 other locations.

PR #246, opened 2024-10-30
Define rootless/non-rootless image tags separately

Secretions:
We currently default to the rootless buildkit image as the only image in the config, and then life-mutate the image tag to what we presume the non-rootless one would be based on a string substitution.

Let's just clearly define the two images and pick the appropriate one.


ddl-kfrench (approved):
🚀

PR #301, opened 2025-02-03
DOM-64953 Adds snapshotter field to helm chart

miguelhar:


deployments/helm/hephaestus/values.yaml
```diff
@@ -47,7 +47,8 @@ newRelic:
 
 # Configuration for buildkit and controller that adds the ability to pull/push images
 # from/to insecure (self-signed TLS) and http registries.
-registries: {}
+registries:
+  {}
```
Secretions, line 51: This is kinda weird

Secretions (approved):

PR #302, opened 2025-02-14
Implement webhook defaulter interface

ddl-dclegg:


ddl-dclegg:
Reimplemented in #305.

PR #307, opened 2025-02-25
Disable healthcheck, fix formatting errs

ddl-dclegg:
### References (optional)

PR #311, opened 2025-03-26
Update outdated packages

ddl-retornam:
Update outdated packages. Upgrade to golang 1.24


ddl-dclegg:
@ddl-retornam I already was trying updates like these out (see #307), but as I was testing things in the catalog, I found that builds just weren't completing. We have to be super careful moving these through. It may be buildkit 0.20.2, may be something else, just haven't pinned it down yet.

ddl-dclegg (approved):

PR #317, opened 2025-04-17
Rootless configs

ddl-dclegg:
- Even when subgid/subuid fixes are baked into the dockerfile, when launched as a pod the istio settings propagate to `/etc/subgid` and `/etc/subuid`. With this configmap update, the right setting (`user:100000:65536`) gets all the way to the container.
- Due to flag discrepancies between docker and containerd, per the [documented tip](https://github.com/moby/buildkit/blob/f8c19098c91a0cd4a2d9cd35e2b3c2a1c8b3f622/docs/rootless.md#change-uidgid) around security-opt flags, we set `--privileged` in rootless too

### DIFF
--- a/deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
+++ b/deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
@@ -50,10 +50,9 @@ spec:
             # NOTE: To change UID/GID, you need to rebuild the image
             runAsUser: {{ .Values.buildkit.rootlessUser }}
             runAsGroup: {{ .Values.buildkit.rootlessUser }}
-            runAsNonRoot: true
-            {{- toYaml .Values.buildkit.rootlessContainerSecurityContext | nindent 12}}
+            {{- toYaml .Values.buildkit.rootlessContainerSecurityContext | nindent 12 }}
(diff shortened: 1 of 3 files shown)


deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
```diff
@@ -50,6 +50,7 @@ spec:
             # NOTE: To change UID/GID, you need to rebuild the image
             runAsUser: {{ .Values.buildkit.rootlessUser }}
             runAsGroup: {{ .Values.buildkit.rootlessUser }}
+            runAsNonRoot: true
```
Secretions, line 53: This should be in in the `rootlessContainerSecurityContext` in values rather than hardcoded here. The only reason the other two are allowed here is just because they have separate variables.

Secretions (approved):

deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
ddl-dclegg: Good call, and I love that you added the values change recently.

Secretions (approved):

PR #320, opened 2025-04-21
fix/debug

ddl-retornam:
- **separate caches for each build**
- **Remove TINI**


.github/actions/push-multiple-container-images/action.yaml
```diff
@@ -32,6 +32,10 @@ inputs:
     description: 'Path to the Dockerfile'
     required: false
     default: 'Dockerfile'
+  suffixe:
```
ddl-dclegg, line 35: I think this should be `suffix`

ddl-dclegg (changes requested):

ddl-dclegg (approved):

PR #336, opened 2025-08-11
PLAT-9527: Add ability to define a secret with google creds

Secretions:
We have a customer who wants to access their google artifact registry from their domino cloud control plane, but because of the way hephaestus works, it will want to do cloud authorization to gcp with these images.

This PR provides a mechanism to setup that authorization to gcr:

```
kubectl -n domino-compute create secret generic gcp-service-account --from-file=key.json=/path/to/service-account/creds.json

# Domino installer config:
[...]
release_overrides:
  chart_values:
     controller:
      manager:
        cloudRegistryAuth:
          gcp:
            mountedCredentials:
              secretName: gcp-service-account
              keyName: key.json
```

This will mount the google credentials into the hephaestus-manager pod, allowing it to auth for these images.

### DIFF
--- a/deployments/helm/hephaestus/Chart.yaml
+++ b/deployments/helm/hephaestus/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 type: application
 name: hephaestus
-version: 0.0.0
-appVersion: "latest"
+version: 0.11.19
+appVersion: "0.11.19"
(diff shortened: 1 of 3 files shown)


ddl-dclegg (approved):
Looks good to me

ldebello-ddl (approved):
LGTM

PR #338, opened 2025-08-14
PLAT-9268 - Switch to socket probe

ddl-dclegg:
Problem:
`hephaestus-buildkit` pods are experiencing cgroup-related probe failures that prevent them from becoming ready, despite the containers starting successfully. 

The liveness and readiness probes are failing with the error:
`rpc error: code = Unknown desc = command error: error opening file '/sys/fs/cgroup/kubepods.slice/kubepods-pod[UUID].slice/crio-[CONTAINER_ID].scope/cgroup.freeze: No such file or directory`

Running `buildctl debug workers` works inside buildkit containers, but is problematic as a probe. Testing the TCP socket is more reliable and not susceptible to cgroup v1/v2 differences as well as others with CRI-O, etc.

Changes:
- Set the `exec.command` to `null`
- Set `tcpSocket.port: 1234` (also set in the buildkit args)
- Adjusted test thresholds

### DIFF
--- a/deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
+++ b/deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
@@ -80,11 +80,9 @@ spec:
               protocol: TCP
               containerPort: {{ .Values.buildkit.service.port }}
           livenessProbe:
-            exec:
-              command:
-                - buildctl
-                - debug
-                - workers
+            exec: null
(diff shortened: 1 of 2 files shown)


ddl-dclegg:
A live openshift deployment had crashlooping buildkit pods, and after this change was applied all pods became healthy and the hung builds completed.

Secretions (approved):

PR #345, opened 2025-09-15
PLAT-9595 - Override persistentVolumeClaimRetentionPolicy defaults

ddl-retornam:
the persistentVolumeClaimRetentionPolicy was added in Kubernetes 1.32.0
the default values are Retain and Retain which leaves [dangling](https://dominodatalab.slack.com/archives/CBXBCC57B/p1757960489754879) 500Gi
volumes behind -> https://dominodatalab.slack.com/archives/CBXBCC57B/p1757960489754879

Upgrading the chart to allow for the values to be overridden

https://dominodatalab.atlassian.net/browse/PLAT-9595


deployments/helm/hephaestus/templates/buildkit/statefulset.yaml
```diff
@@ -8,6 +8,11 @@ spec:
   serviceName: {{ include "hephaestus.buildkit.fullname" . }}
   podManagementPolicy: Parallel
   replicas: {{ .Values.buildkit.replicaCount }}
+  {{- if semverCompare ">=1.32.0" .Capabilities.KubeVersion.Version }}
+  persistentVolumeClaimRetentionPolicy:
+    whenDeleted: {{ .Values.buildkit.persistence.pvcRetentionPolicy.whenDeleted | default "Delete" }}
+    whenScaled: {{ .Values.buildkit.persistence.pvcRetentionPolicy.whenScaled | default "Retain" }}
```
Secretions, line 14: I don't think we need a default here, since it's already in the values.

Secretions (approved):

PR #357, opened 2025-10-02
Build updates

ddl-dclegg:


ddl-dclegg:
Largely superceded by #382, can bring back [small set](https://github.com/dominodatalab/hephaestus/commit/434d5a9b2bb0ea1fb6a1e9df88c430c1cb28d950) of changes after that merges

PR #379, opened 2026-03-26
PLAT-9845 Add jobNodeSelector to CRD hook jobs

miguelhar:
Add jobNodeSelector to the crd-apply and crd-delete hook Jobs so they can target the domino-platform-jobs nodepool, falling back to controller.nodeSelector if unset.

### DIFF
--- a/deployments/helm/hephaestus/templates/crd-hooks.yaml
+++ b/deployments/helm/hephaestus/templates/crd-hooks.yaml
@@ -99,7 +99,7 @@ spec:
           env:
             {{- toYaml . | nindent 12 }}
           {{- end }}
-      {{- with .Values.controller.nodeSelector }}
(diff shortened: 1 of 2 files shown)


ddl-dclegg (approved):
Looks good, just one question -- are `tolerations` and `affinity` in the same boat for jobs? Should we be handling them similarly?

miguelhar:
> Looks good, just one question -- are `tolerations` and `affinity` in the same boat for jobs? Should we be handling them similarly?

I would say no for this usecase, we are moving those workloads that currently just have a nodeselector constraint to use the jobNodeSelector, If a job/cronjob has co-location constraints that specify where it needs to run based on another workload, (i.e backup job needs to run in the git node) then its not a candidate for this change. taints remain unaffected and applicable

PR #417, opened 2026-06-29
Add hephaestus_imagebuild_phase_total metric (DOM-78643)

alexglenn-ddl:
## Context

Pillar 1 of the OT-3690 build-reliability work (quay.io outage — image builds down ~16h, **no alert fired**).

Today Hephaestus emits **no build-outcome signal it owns**:
- Build SLOs/dashboards depend on the New Relic `BuildStatusChange` custom event, which Hephaestus does **not** emit — it's derived downstream from the AMQP transition message. If that consumer breaks, build observability silently goes green.
- The structured `error.class` **never crosses AMQP** (only a free-text `ErrorMessage` does), so a faceted, reason-labeled failure signal is impossible to recover downstream.

This PR adds that signal at the source.

## Change

New Prometheus counter emitted directly by the controller at terminal phase transitions:

```
hephaestus_imagebuild_phase_total{phase, failure_reason}
```

- `phase`: `Succeeded` | `Failed`.
- `failure_reason`: empty on success; on failure it reuses the **same vocabulary as the New Relic `error.class`** (`CredentialsValidateError`, `ImageBuildError`, `WorkerLeaseError`, `ClusterSecretsReadError`, `CredentialsPersistError`, `WorkerClientInitError`, plus `NotRunning` for the controller-restart recovery path) — so the metric correlates 1:1 with APM `TransactionError` facets.

### Implementation notes
- Registered on controller-runtime's global `metrics.Registry`, so it's auto-served on the existing `/metrics` endpoint — no new wiring.
- Increments go through thin `failBuild`/`succeedBuild` wrappers in the dispatcher; the shared `phase.TransitionHelper` is **unchanged** (imagecache and any future phased CRs are unaffected). `failure_reason` is captured at the dispatcher call sites because it is not available inside `phase.go`.
- Terminal phases short-circuit the reconcile entry, so each build increments once on the normal path.

### Consumers (separate work, TerraMonitor)
- Faceted `error.class`/`failure_reason` alert (cause-layer detection).
- Demand-vs-completion + loss-of-signal (effect-layer detection).
- Completion-based SLI for reporting (non-paging).

## Testing
- `go build ./...`, `go vet ./pkg/controller/imagebuild/...`, `go test ./pkg/controller/imagebuild/...` all pass.
- New `metrics_test.go` asserts label wiring/values via `testutil.CollectAndCompare`.

## Follow-up (not in this PR)
- Confirm `manager.metricsAddr` binds `0.0.0.0:8080`, not loopback — otherwise the `/metrics` endpoint (including existing controller-runtime metrics) isn't scrapeable. ServiceMonitor only needed if the bind is non-loopback and discovery still misses it.

Refs: OT-3690, DOM-78643.

### DIFF
--- a/pkg/controller/imagebuild/component/builddispatcher.go
+++ b/pkg/controller/imagebuild/component/builddispatcher.go
@@ -277,25 +277,42 @@ func (c *BuildDispatcherComponent) Reconcile(coreCtx *core.Context) (ctrl.Result
 		populateBuildStatus(obj, buildLog, img, imageName)
 	}
 
-	c.succeedBuild(coreCtx, obj)
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, c.succeedBuild(coreCtx, obj)
 }
 
-// failBuild records a terminal Failed transition for the build, incrementing the
-// phase metric with the given reason before delegating to the phase helper. reason
-// mirrors the New Relic error.class set at the call site.
+// failBuild records a terminal Failed transition for the build and increments the phase
+// metric with the given reason. The metric is incremented only after the transition is
+// durably persisted, so a reconcile requeue triggered by a failed status write does not
+// double-count (the retry re-persists and increments exactly once). reason mirrors the
+// New Relic error.class set at the call site. On a successful persist the original build
+// error is returned unchanged, preserving the existing reconcile requeue/error behavior.
 func (c *BuildDispatcherComponent) failBuild(
 	ctx *core.Context, obj *hephv1.ImageBuild, err error, reason string,
 ) error {
+	if persistErr := c.phase.SetFailed(ctx, obj, err); persistErr != nil {
+		// Terminal phase not persisted; do not count. The non-nil return requeues the
+		// reconcile, which retries the write and increments once it lands.
+		return persistErr
+	}
+
 	recordImageBuildPhase(hephv1.PhaseFailed, reason)
-	return c.phase.SetFailed(ctx, obj, err)
+
+	return err
 }
 
-// succeedBuild records a terminal Succeeded transition for the build, incrementing the
-// phase metric (with an empty failure_reason) before delegating to the phase helper.
-func (c *BuildDispatcherComponent) succeedBuild(ctx *core.Context, obj *hephv1.ImageBuild) {
+// succeedBuild records a terminal Succeeded transition for the build and increments the
+// phase metric (with an empty failure_reason). As with failBuild, the metric is
+// incremented only after the transition is durably persisted; a failed status write is
+// returned so the reconcile requeues and retries rather than silently counting a
+// non-durable transition.
+func (c *BuildDispatcherComponent) succeedBuild(ctx *core.Context, obj *hephv1.ImageBuild) error {
+	if persistErr := c.phase.SetSucceeded(ctx, obj); persistErr != nil {
+		return persistErr
+	}
+
 	recordImageBuildPhase(hephv1.PhaseSucceeded, "")
-	c.phase.SetSucceeded(ctx, obj)
+
+	return nil
 }
 
 func (c *BuildDispatcherComponent) processCancellations(log logr.Logger) {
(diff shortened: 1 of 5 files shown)


alexglenn-ddl:
**Heads-up for review — `phase.TransitionHelper` return-semantics change (513c3f4).**

The double-count fix required surfacing the status-write result that `updateStatus` was silently swallowing, so a few signatures in the shared `phase` package changed:

- `updateStatus` now **returns** its error (still logs + emits the `StatusUpdate` Event as before).
- `SetSucceeded` gains an `error` return (was `void`).
- `SetFailed` now returns the **status-write error** (nil on success) rather than the passed-in build error. `err` is still used to populate the failure condition message.

Why this is safe:

- The only live callers of `SetFailed`/`SetSucceeded` are the `failBuild`/`succeedBuild` wrappers in `builddispatcher.go` — the imagecache callers in `cachewarmer.go` are all commented out — so those wrappers fully own the interpretation of the new return value.
- The reconcile still returns non-nil on a build failure (the original build error on a durable persist, the persist error otherwise), so `controller_runtime_reconcile_errors_total` behavior is unchanged.
- Doc comments on both `SetSucceeded`/`SetFailed` spell out the new contract, so it doesn't read as a surprise when viewing `phase.go` in isolation.

The regression tests (`TestFailBuildCountsOnlyOnDurableTransition` / `TestSucceedBuildCountsOnlyOnDurableTransition`) drive a flaky status write and assert exactly-once counting across requeues; they fail on the pre-fix increment ordering and pass with the fix.

pkg/controller/imagebuild/component/builddispatcher.go
```diff
@@ -276,8 +277,42 @@ func (c *BuildDispatcherComponent) Reconcile(coreCtx *core.Context) (ctrl.Result
...
+func (c *BuildDispatcherComponent) failBuild(
+	ctx *core.Context, obj *hephv1.ImageBuild, err error, reason string,
+) error {
+	if persistErr := c.phase.SetFailed(ctx, obj, err); persistErr != nil {
+		// Terminal phase not persisted; do not count. The non-nil return requeues the
+		// reconcile, which retries the write and increments once it lands.
+		return persistErr
+	}
+
+	recordImageBuildPhase(hephv1.PhaseFailed, reason)
```
ddl-bira-ignacio, line 298: question: if the cache is stale and the status is still Running/Initializing (which can happen) then could we be generating two metrics? one with real reason and the other with NotRunning? 

Probably hard to work around, the real solution, if one is needed, would be for the cache to be up to date (something we don't really control at the moment)

ddl-bira-ignacio (approved):

pkg/controller/imagebuild/component/builddispatcher.go
alexglenn-ddl: Good thing to probe — I checked, and this case is actually closed by optimistic concurrency, so it holds even with a stale cache.

Walkthrough: when the first reconcile persists `Failed` (with the real reason), the API server bumps the object's `resourceVersion`. A later reconcile that reads a **stale** cache still holds the *old* resourceVersion and phase `Running`, so it enters the `NotRunning` branch and calls `Status().Update` — but that write carries the stale resourceVersion and the API server rejects it with a **409 Conflict** (`object was modified`). Because the counter now increments *only after a successful write*, the `NotRunning` increment never happens. Once the cache catches up, the reconcile hits the terminal `Succeeded/Failed` short-circuit and also doesn't increment.

So the invariant is: at most one status write can win the transition to a terminal phase (optimistic concurrency), and only the winning write increments → exactly once, correct reason. We don't need the cache to be fresh for the metric to be correct.

Added a regression test for exactly this (445a585, `TestStaleCacheDoesNotDoubleCount`): a stale copy at the Running resourceVersion, real reason persisted first, then the stale `NotRunning` write returns `IsConflict=true` and the counter stays `{real=1, notrunning=0}`.

The one case where the reason is degraded is when the *first* write fails to persist at all — then a requeue records the outcome once as `NotRunning`. That's a single count with a lossy reason (not a double count), and it's the documented residual: perfect reason attribution would require persisting the reason, which is the same gap this metric is working around.

ddl-bira-ignacio (approved):

PR #420, opened 2026-07-07
DOM-78638: don't fail builds when an unused registry is down

miguelhar:
Jira: https://dominodatalab.atlassian.net/browse/DOM-78638

Reference: https://dominodatalab.atlassian.net/browse/OT-3690

## Problem
Before a build starts, Hephaestus checks credentials for every configured registry, and one failed check fails the build. During the quay.io outage (OT-3690) that took down Cloud builds that never touch quay.io. The creds were valid. quay.io was unreachable, and the retry burned 31 seconds before failing the build.

## Change
`credentials.Verify` follows two rules:

1. Only creds for the registries a build **targets or uses for caching** are checked: the registries in `spec.images` and `spec.importRemoteBuildCache`, the refs Hephaestus has in hand. A base image pull is a use and is not covered. Base images live in the Dockerfile, resolving them means parsing it, and buildkit authenticates those itself. If a ref cannot be parsed the registry is unknown, so every cred is checked.
2. A checked registry that never answered is skipped with a warning. Anything it did answer with (401, 403, 5xx) is a credential problem and fails the build before a worker is leased. The first answer sticks: a registry that rejects the cred and then goes down mid retry still fails the build. A cancelled build context stays fatal and keeps the credential failures already found.

```
                     ┌──────────────────────────────────────────────────────────┐
                     │  ImageBuild dispatched with creds for:                   │
                     │  quay.io · ECR · docker.io      build pushes to ECR only │
                     └──────────────────────────────────────────────────────────┘

  BEFORE                                      AFTER
  ══════                                      ═════
  validate ALL configured registries         validate ONLY target + cache registries
        │                                           │
        ├─ quay.io outage                           ├─ quay.io outage ..... not checked
        │    retry ~31s                             ├─ quay.io bad cred ... not checked
        │    ✖ BUILD FAILS                          │
        │                                           ├─ ECR bad cred ....... ✖ fail fast (kept)
        ├─ quay.io bad cred                         ├─ ECR 403 ............ ✖ fail fast (kept)
        │    ✖ BUILD FAILS                          ├─ ECR unreachable .... warn + continue
        │                                           │     buildkit fails the push
        ▼                                           ▼     itself if ECR stays down
   build never starts                          build runs
```

| Scenario | Before | After |
|---|---|---|
| Unused registry down **([OT-3690](https://dominodatalab.atlassian.net/browse/OT-3690))** | build fails after a 31 second retry | ignored, build runs |
| Unused registry bad cred | build fails | ignored, build runs |
| Base image registry bad cred | build fails at validation, clear message | build fails at pull with buildkit's error, after a worker is leased |
| Target or cache registry bad cred | build fails fast | build fails fast (unchanged) |
| Target or cache registry returns 403 | build fails | build fails (unchanged) |
| Target or cache registry rejects the cred, then goes down | build fails | build fails (unchanged) |
| Target or cache registry never answers | build fails after retry | warning logged, buildkit surfaces the real push error |
| Build context cancelled | build fails | build fails (unchanged) |

Only a registry that did not respond at all is skipped. A bad cred is a bad cred whatever status it comes back with, and `errdefs.IsUnauthorized` cannot be the gate: docker only produces it for a 401 body with no JSON content type, which real registries never send, so a bearer 401 from quay.io arrives as a plain `*url.Error` like any transport failure. Gating on it would let a wrong password through to a leased worker.

## Test
`TestVerify`, 16 subtests: unreachable registry is skipped, Basic and Bearer 401 and 403 are fatal, a skipped registry does not hide a bad cred elsewhere, a rejection followed by an outage still fails, cancelled context is fatal, unused registry is never contacted, cache import registry is checked, legacy Docker Hub keys match hub images, host case and URL userinfo in a server key are normalised away, ports are part of the host in both directions, an unparseable ref checks everything, an empty ref list checks everything.

Verified on cluster `mhhepha138692`, twice: unused registry with a bad cred builds, bad bearer cred fails before a worker is leased, unreachable target registry is skipped and buildkit reports the push failure. Logs and manifests in the [evidence comment](https://github.com/dominodatalab/hephaestus/pull/420#issuecomment-5283034914).

### DIFF
--- a/pkg/controller/support/credentials/credentials.go
+++ b/pkg/controller/support/credentials/credentials.go
@@ -3,6 +3,7 @@ package credentials
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -11,6 +12,7 @@ import (
 
 	"github.com/containerd/errdefs"
 
+	"github.com/distribution/reference"
 	typesregistry "github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/registry"
 	"github.com/go-logr/logr"
@@ -132,7 +134,44 @@ func Persist(
 	return dir, helpMessage, err
 }
 
-func Verify(ctx context.Context, configDir string, insecureRegistries []string, helpMessage []string) error {
+// normalizeRegistryHost reduces a docker config server key or an image ref
+// domain to a plain hostname so the two can be compared. Docker Hub goes by
+// several names, so they all collapse to "docker.io".
+func normalizeRegistryHost(server string) string {
+	host := registry.ConvertToHostname(server)
+	if host == "index.docker.io" || host == "registry-1.docker.io" {
+		host = "docker.io"
+	}
+
+	return host
+}
+
+// usedRegistryHosts returns the set of registry hostnames the build is known
+// to use, taken from the image refs it will push.
+func usedRegistryHosts(images []string) map[string]bool {
+	hosts := map[string]bool{}
+	for _, image := range images {
+		named, err := reference.ParseNormalizedNamed(image)
+		if err != nil {
+			// Image refs are validated by the webhook. A ref that does not
+			// parse cannot match any registry, so it adds nothing here.
+			continue
+		}
+
+		hosts[normalizeRegistryHost(reference.Domain(named))] = true
+	}
+
+	return hosts
+}
+
+func Verify(
+	ctx context.Context,
+	logger logr.Logger,
+	configDir string,
+	insecureRegistries []string,
+	helpMessage []string,
+	images []string,
+) error {
 	filename := filepath.Join(configDir, "config.json")
 	data, err := os.ReadFile(filename)
 	if err != nil {
@@ -149,25 +188,57 @@ func Verify(ctx context.Context, configDir string, insecureRegistries []string, 
 		return err
 	}
 
+	usedHosts := usedRegistryHosts(images)
+
 	var errs []error
 	for server, auth := range configJSON.Auths {
+		// Only check credentials for registries this build pushes to. A cred
+		// for any other registry cannot fail the build here, no matter its
+		// state. If no images are given, check everything.
+		if len(usedHosts) > 0 && !usedHosts[normalizeRegistryHost(server)] {
+			logger.Info("Skipping credential check for registry not used by this build", "registry", server)
+			continue
+		}
+
 		auth.ServerAddress = server
 
+		var authErr error
 		err := wait.ExponentialBackoffWithContext(ctx, defaultBackoff, func(ctx context.Context) (bool, error) {
-			if _, _, err = svc.Auth(ctx, &auth, "DominoDataLab_Hephaestus/1.0"); err != nil {
-				if errdefs.IsUnauthorized(err) {
-					return false, err
+			if _, _, authErr = svc.Auth(ctx, &auth, "DominoDataLab_Hephaestus/1.0"); authErr != nil {
+				if errdefs.IsUnauthorized(authErr) {
+					return false, authErr
 				}
 				return false, nil
 			}
 
 			return true, nil
 		})
-		if err != nil {
+		if err == nil {
+			continue
+		}
+
+		// A cancelled or expired build context is fatal. It is not a registry
+		// being unreachable, so do not skip it.
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			return err
+		}
+
+		// Bad credentials fail the build now so the error is clear and no worker
+		// time is wasted on creds we know are wrong.
+		if errdefs.IsUnauthorized(err) {
 			//nolint:lll
 			detailedErr := fmt.Errorf("client credentials are invalid for registry %q.\nMake sure the following sources of credentials are correct: %s.\nUnderlying error: %w", server, strings.Join(helpMessage, ", "), err)
 			errs = append(errs, detailedErr)
+			continue
 		}
+
+		// A registry we cannot reach must not fail a build that does not use it.
+		// Buildkit still fails the build later if it actually needs this registry.
+		reason := err
+		if authErr != nil {
+			reason = authErr
+		}
+		logger.Info("Skipping credential check for unreachable registry", "registry", server, "reason", reason.Error())
 	}
 	if len(errs) != 0 {
 		return multierr.Combine(errs...)
--- a/pkg/controller/support/credentials/credentials.go
+++ b/pkg/controller/support/credentials/credentials.go
@@ -180,7 +180,6 @@ func targetedRegistryHosts(refs []string) map[string]bool {
 // *url.Error, which satisfies net.Error whatever the cause, and docker's bearer
 // transport returns registry status errors through it. Gating on that reads a
 // bearer 401 as an outage and lets a typo'd password reach a leased worker.
-// Every registry we care about speaks bearer.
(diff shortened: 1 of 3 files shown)


ddl-dclegg (approved):

miguelhar:
Unrelated to this PR, but spotted while tracing `spec.importRemoteBuildCache` for the credential-check change: `pkg/buildkit/buildkit.go:265-274` **assigns** `solveOpt.CacheImports` inside its loop instead of appending, so only the last ref survives.

```go
for _, ref := range opts.ImportCache {
    solveOpt.CacheImports = []bkclient.CacheOptionsEntry{
        {Type: "registry", Attrs: map[string]string{"ref": ref}},
    }
}
```

So an ImageBuild listing more than one cache ref silently imports only one. Pre-existing and out of scope here — filing separately.

miguelhar:
## Verified on a live cluster

| | |
|---|---|
| Deployment | `mhhepha138692` (`dev-aws-eks`, us-west-2) |
| Installed version | `6.4.0+mh-DOM-78638-hephaestus-pr-420-2`, from cerebrotech/catalog#6590 |
| Running image | `670838783587.dkr.ecr.us-west-2.amazonaws.com/domino/hephaestus:pr-420` |
| Commit under test | `f0f0724`, from the quay `pr-420` label `org.opencontainers.image.revision` |

### Results

| Scenario | Outcome | Worker leased | Only this build does it |
|---|---|---|---|
| Unused registry, **bad credential** | **Succeeded**, registry never contacted | yes | `0.12.18` fails with a quay 401 |
| Target **bearer** registry, bad credential | **Failed** before dispatch | **no** | `5516ee2` skips it and proceeds |
| Target registry **unreachable** | Failed at push, not at validation | yes | `0.12.18` fails at validation |
| 6 real environment builds | all **Succeeded** | yes | happy path unchanged |

Rows 1 and 2 are the pair that matters. An unused 401 is tolerated, a targeted one is not, so neither can be mistaken for the other.

### 1. Unused registry, bad credential (the OT-3690 case)

`images` pushes to ECR only. `quay.io` carries a wrong password and is never referenced by the build.

```
15:09:52.339419  Validating registry credentials  {"logKey":"dom78638-unused-badcred"}
15:09:52.339473  Skipping credential check for registry not used by this build  {"registry":"quay.io"}
```

54µs apart: quay.io was skipped, not attempted and tolerated. `Succeeded`, `buildTime=1.451s`, `digest=sha256:37aa3b17dea5fc6cafe4cbf7046fc93f26e1307e22702aa07c0761dbd77d2bcc`, no `client credentials are invalid` anywhere in the log. On `main` this build fails.

### 2. Target bearer registry, bad credential

```
client credentials are invalid for registry "quay.io".
Make sure the following sources of credentials are correct: basic authentication username and password.
Underlying error: Get "https://quay.io/v2/": unauthorized: Invalid Username or Password
```

`Initializing` to `Failed`, `builderAddr` empty, no worker leased. It took 33 seconds, the full backoff: docker parses quay's JSON 401 body as `errcode.Errors`, which `errdefs.IsUnauthorized` does not recognise, so the short circuit in the retry loop never fires against a real registry.

That underlying error is why `f0f0724` exists. `http.Client` wraps every failure in `*url.Error`, which satisfies `net.Error` whatever the cause, and docker's bearer transport returns registry status errors through it. Measured from `registry.Service.Auth`:

| Flow | Error chain | `net.Error` |
|---|---|---|
| bearer 401 | `registry.unauthorizedErr` around `*url.Error` around `errcode.Error` | true |
| bearer 403 | `*url.Error` around `errcode.Error` | true |
| basic 401 and 403 | plain error | false |

Testing the top level error read a bad bearer credential as an outage and let it through to a leased worker. `unreachable()` now unwraps `*url.Error` and tests the transport error underneath. Two subtests fail without that change.

### 3. Target registry unreachable

```
Skipping credential check for unreachable registry
  {"registry":"dom78638-nonexistent-registry.invalid",
   "reason":"Get \"https://…/v2/\": dial tcp: lookup … on 172.20.0.10:53: no such host"}
```

Reached `Running` with a leased worker, built the image, and failed at push with buildkit's own DNS error. A DNS failure is unreachable, a quay 401 is fatal, both against real registries.

### Identifying the commit

`hephaestus-controller version` prints `dev`. The Dockerfile passes `-X main.Version` but `Version` lives in `pkg/cmd/controller` and `main.go` never assigns it, so the flag does nothing. Identification is behavioural, through log lines and outcomes no other version produces. Filing the `main.Version` gap separately.

### Reproducing

Apply against any deployment running this build. Scenario 1 is the OT-3690 case. Swap the ECR ref for the target cluster's own.

```yaml
apiVersion: hephaestus.dominodatalab.com/v1
kind: ImageBuild
metadata: {name: dom78638-unused-badcred, namespace: domino-compute}
spec:
  logKey: dom78638-unused-badcred
  dockerfileContents: |
    FROM public.ecr.aws/docker/library/alpine:3.20
    RUN echo dom78638-verify
  images:
    - <acct>.dkr.ecr.<region>.amazonaws.com/<deployment>/environment:dom78638-verify
  registryAuth:
    - server: <acct>.dkr.ecr.<region>.amazonaws.com   # used, cloud auth
    - server: quay.io                                 # UNUSED, deliberately bad
      basicAuth: {username: bogus, password: bogus}
```

Expect `Succeeded` plus a skip line naming quay.io. Scenario 2 is the same manifest with `images` pointed at `quay.io/domino/dom78638-verify:test`; expect `Failed` with an empty `builderAddr`. Scenario 3 points `images` and `registryAuth.server` at a host that does not resolve; expect a leased worker and a failure at push.

### After this run

- 453875c reworded the skip line to `Skipping credential check: registry is not a build target or cache ref`. Same condition, same code path.
- 8bc9672 keeps the first answer a registry gives, so a target that rejects the cred and then goes down still fails the build. Staging that needs a fake registry with a valid TLS cert, or an operator config change and restart, so it is covered by the `rejection_is_not_erased_by_a_later_outage` unit test, which asserts the outage happened and fails with the fix reverted.
- Scenarios 1 to 3 were rerun on the same deployment with identical results. Manager logs and status YAML are captured and available on request.

pkg/controller/support/credentials/credentials.go
```diff
@@ -149,25 +223,72 @@ func Verify(ctx context.Context, configDir string, insecureRegistries []string,
 		return err
 	}
 
+	targetedHosts := targetedRegistryHosts(refs)
+	if targetedHosts == nil {
+		logger.Info("Checking every credential: a build reference could not be parsed", "references", refs)
+	}
```
ddl-dclegg: Not-blocking: with the new signature we could have no refs
```suggestion
	if len(refs) == 0 || targetedHosts == nil {
		logger.Info("Checking every credential: no build targets to scope to, or a reference could not be parsed", "references", refs)
	}
```

ddl-dclegg (approved):
Looks good. My gut says we should be adding in cert handling since we're talking about remotes. If I remember correctly, a TLS handshake failure happens before any HTTP response, but we should treat it like one. Maybe at least
`x509.CertificateInvalidError`, `x509.UnknownAuthorityError`, `x509.HostnameError`.

Probably a few more additions regarding `ImportRemoteBuildCache` too since we're lumping it with `Images` in `credentials.Verify`

pkg/controller/support/credentials/credentials.go
miguelhar: Good catch. `images` is `omitempty` with no `minItems`, so an empty ref list is a real case. Applied as suggested in d1a5f91.

PR #435, opened 2026-08-29
VUL-1647: patch amqp091-go and fix build checks that assume a shell in distroless

miguelhar:
Jira: https://dominodatalab.atlassian.net/browse/VUL-1647

**amqp091-go 1.11.0 to 1.13.0** — CVE-2026-79921: a malicious broker can push body frames past the negotiated `frame_max` and the client allocates for them. Indirect through `dominodatalab/amqp-client`. Neither scanner flags it (still `Received` in NVD, no OSV or GitHub advisory), so `go version -m` on `sha-76eaf7c` is the evidence we ship 1.11.0.

**x/crypto to v0.55.0** for CVE-2026-56854, which the first pass missed.

**Verify Build container commands** — failed on all 315 runs since #382. Three checks ask the images for things they don't have: the amd64/arm64 shell probes run `--entrypoint /bin/sh` against `distroless/static:nonroot`, which has no shell, and the debug check passes `bash` as an argument to the controller entrypoint instead of replacing it. The images are fine; the probes were wrong. Shell probe now calls the entrypoint directly, debug check uses `--entrypoint bash`.

Scanned the built amd64 image with grype and trivy: all 23 ticket CVEs gone. Seven advisories remain that no ticket tracks. Six have no published fix. The seventh, CVE-2026-33997, wants docker/docker 29.3.1 against our 28.5.2 — a major jump on a buildkit dependency, so it needs its own ticket rather than riding this one.

Green on all five jobs: https://github.com/dominodatalab/hephaestus/actions/runs/33258223060


miguelhar:
Ran the patched workflow against the current main image before merge: https://github.com/dominodatalab/hephaestus/actions/runs/33258223060

All five jobs green on `sha-76eaf7c`. Both architectures print `OCI image build controller using buildkit` from the entrypoint, and the debug image reaches bash. That is the first green run this workflow has ever had (315 failures before it).

ddl-r-abdulaziz (approved):

PR #440, opened 2026-09-10
DOM-78638: fail fast on a bad registry cert instead of burning the retry backoff

miguelhar:
Jira: https://dominodatalab.atlassian.net/browse/DOM-78638

Follow-up to #420: dclegg's review comment on that PR landed after merge and was never addressed.

## Problem
A registry with a bad cert (unknown authority, expired, wrong host) unwraps to `*tls.CertificateVerificationError`, which matches neither `errdefs.IsUnauthorized` nor `net.Error`. `unreachable()` correctly treats it as answered rather than skipping it, but the retry loop's only fast exit checks `IsUnauthorized`, so a bad cert burns the full ~31s backoff before the build fails, the same latency #420 removed for other cases.

## Change
Add `certRejected`, matching `*tls.CertificateVerificationError`, and short-circuit the retry loop on it alongside `IsUnauthorized`.

## Test
`TestCertRejected` exercises the real error `http.Client` produces against an untrusted cert. `registry.NewService` treats `127.0.0.0/8` as insecure by default and skips cert verification there, so this can't be reproduced through `Verify()` with a local listener; `certRejected` is tested directly instead.

### DIFF
--- a/pkg/controller/support/credentials/credentials.go
+++ b/pkg/controller/support/credentials/credentials.go
@@ -2,6 +2,7 @@ package credentials
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


ddl-dclegg (approved):
Looks good! `*tls.CertificateVerificationError` catches `x509.UnknownAuthorityError`, `x509.HostnameError`, `x509.CertificateInvalidError`, and anything else the handshake wraps. 👏 

On the `ImportRemoteBuildCache` half of my #420 comment, I now see that `Images` and `ImportRemoteBuildCache` are merged into one host list before `Verify` runs, so this fast-exit applies to cache-only hosts. 

:shipit:

PR #441, opened 2026-09-11
PLAT-9571: lazy AWS config load so ECR auth survives slow networking

miguelhar:
Jira: https://dominodatalab.atlassian.net/browse/PLAT-9571

`ecr.Register` called `config.LoadDefaultConfig` once at startup and on failure skipped registering the loader. Under istio ambient the pod has no routable network for its first seconds (DOM-70981), IMDS times out at 5s, and ECR auth is gone until the pod restarts.

The loader is now registered unconditionally and the AWS config loads lazily, memoized on success. A failed startup load surfaces as an error from `authenticate`, which `RefreshingAuthProvider` already retries per build.

Behavior change only when the load fails: a build that references a private ECR host with no static secret on a non AWS cluster now fails after the provider's retries instead of instantly, and logs the load error.

### DIFF
--- a/pkg/controller/support/credentials/cloudauth/ecr/ecr.go
+++ b/pkg/controller/support/credentials/cloudauth/ecr/ecr.go
@@ -59,7 +59,7 @@ func Register(ctx context.Context, logger logr.Logger, registry *cloudauth.Regis
 	// Pod networking may not be up yet (istio ambient, DOM-70981). Registering
 	// unconditionally lets authenticate finish the load on the first build.
 	if err := loadConfig(ctx, logger); err != nil {
-		logger.Info("ECR registered, AWS config load deferred", "error", err)
+		logger.Info("ECR registered, AWS config load deferred to first use", "error", err)
 		return nil
 	}
 
--- a/pkg/controller/support/credentials/cloudauth/ecr/ecr.go
+++ b/pkg/controller/support/credentials/cloudauth/ecr/ecr.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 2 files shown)


pkg/controller/support/credentials/cloudauth/ecr/ecr.go
```diff
@@ -91,7 +118,11 @@ func authenticate(ctx context.Context, logger logr.Logger, url string) (*registr
 		return nil, err
 	}
 
-	client := newClient(match[urlRegexRegionIndex])
+	client, err := newClient(ctx, logger, match[urlRegexRegionIndex])
+	if err != nil {
+		logger.Info(err.Error())
```
ddl-r-abdulaziz:
```suggestion
		logger.Info("could not authenticate", "error", err)
```

ddl-r-abdulaziz (approved):

pkg/controller/support/credentials/cloudauth/ecr/ecr.go
```diff
@@ -66,19 +90,22 @@ func Register(ctx context.Context, logger logr.Logger, registry *cloudauth.Regis
...
 	return nil
 }
 
-func newECRClient(region string) ecrClient {
+func newECRClient(ctx context.Context, logger logr.Logger, region string) (ecrClient, error) {
+	if err := loadConfig(ctx, logger); err != nil {
+		return nil, fmt.Errorf("failed to load AWS config: %w", err)
+	}
+
 	c := awsConfig
```
ddl-r-abdulaziz: this is threadsafe, but it's not immediately obvious why. Perhaps it would feel safer upon inspection if loadConfig returned the config instead.

pkg/controller/support/credentials/cloudauth/ecr/ecr.go
miguelhar: Done

pkg/controller/support/credentials/cloudauth/ecr/ecr.go
miguelhar: Done

ddl-r-abdulaziz:
LGTM, tested well in my load test.

